PR #39, opened 2014-12-18
safe_string

vbmithr:


lib/cstruct.mli
```diff
@@ -244,8 +244,11 @@ val blit_from_string: string -> int -> t -> int -> int -> unit
     valid substring of [src], or if [dstoff] and [len] do not
     designate a valid segment of [dst]. *)
 
-val blit_to_string: t -> int -> string -> int -> int -> unit
-(** [blit_to_string src srcoff dst dstoff len] copies [len] characters
+val blit_from_bytes: Bytes.t -> int -> t -> int -> int -> unit
+(** See {!blit_from_string}. *)
+
+val blit_to_bytes: t -> int -> Bytes.t -> int -> int -> unit
```
avsm: This breaks the interface substantially and should go into Cstruct 2.0; e.g. shared-memory-ring fails

```
# File "lib/ring.ml", line 407, characters 4-26:
# Error: Unbound value Cstruct.blit_to_string
# Hint: Did you mean blit_from_string?
```

avsm: Or we could leave `blit_to_string` as an alias to `blit_to_bytes` and mark it deprecated.

samoht:
Can you revert the change in the external API? I'm happy to merge the simple string -> byte change otherwise.

vbmithr:
I think it's better to leave it for when this package will be properly safe_string compliant.
`blit_to_string` cannot exist in a safe_string world. I'm closing this for now.

PR #40, opened 2015-01-15
Add buffer function

talex5:
This is useful if you want to buffer data into fixed-sized chunks.
It was originally in mirage-net-xen. @samoht suggested moving it here (https://github.com/mirage/mirage-net-xen/pull/17/files#r22857225).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+* Added `buffer` function.
+
 1.5.0 (2014-11-24):
 * Make `camlp4` an optional build-time dependency (#35).
 * Remove `ounit` as a dependency in the `opam` file.
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -231,6 +231,22 @@ let copyv ts =
     ) 0 ts in
   dst
 
+let buffer src dst =
+  let rec aux dst n = function
+    | [] -> n, []   (* src exhausted *)
+    | hd::tl ->
+        let avail = len dst in
+        let first = len hd in
+        if first <= avail then (
+          blit hd 0 dst 0 first;
+          aux (shift dst first) (n + first) tl
+        ) else (
+          blit hd 0 dst 0 avail;
+          let rest_hd = shift hd first in
+          (n + avail, rest_hd :: tl)    (* dst full *)
+        ) in
+  aux dst 0 src
+
 let to_string t =
   let sz = len t in
   let s = String.create sz in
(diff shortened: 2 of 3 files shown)


samoht:
I've maybe merged that a bit quickly. I'm not a bit fan of the name, as `Cstruct.buf` already refers to the underlying bigarray buffer.

samoht:
Sorry for the mess. I'd prefer to merge #41 instead. @talex5  and @avsm any though?

dbuenzli:
`blitv` seems more consistent with the exisiting `copyv`, `lenv` (what does `v` stand for though ? I would have used `l`). However the doc string of #40 is better: to the point, use of present tense and no need to mention exceptions if a function doesn't raise, it's the raising function that should always document the fact they may do so.

talex5:
I changed the name from `blitv` because it's not very similar to the `blit` function (which copies everything or fails). "Buffer" as a verb describes what it does quite well, but it could indeed be read as a noun. Suggestions welcome, though.

dbuenzli:
`fillv` ?

talex5:
The only problem with `fillv` is that the argument order feels wrong. `buffer x y` suggests that `x` is the thing being buffered (into `y`), but `fill x y` suggests that `x` is being filled. Reversing the order would be inconsistent with `blit` though. Maybe `fill_fromv`?

I don't feel strongly about it though. The argument types are different, so they can't be mixed up.

dbuenzli:
I wouldn't care about consistency with `blit` and switch the argument order (which would yield a better order for partial application IMHO).

avsm:
I'm inclined to label the arguments to make the code really explicit about `~src ~dst` rather than risk any confusion.  How about using a name similar to Lwt and going with:

```
val transfer : src:t list -> dst:t -> int * t list
```

talex5:
If you want to support partial application, it would be better to leave one unlabelled. Maybe `fillv ~dst src`? The problem with `transfer` is it suggests mutating the source...

avsm:
I'm really not that bothered about partial application in Cstruct -- it's far better to be explicit with this sort of code since we're dealing with untyped byte buffers.  I'd be happy with `fillv ~src ~dst`... this does support partial application more flexibly than without the label, but with a bit more verbosity.

avsm:
In the absence of any further comments, I'm going to go with `fillv ~src ~dst`

samoht:
Good for me

talex5:
Fine with me.

PR #41, opened 2015-01-23
Add blitv function

avsm:
```
val blitv: t list -> t -> int * t list
(** [blitv src dst] will copy the list of [src] buffers into [dst] and
    return a tuple of the total number of bytes written and a list of
    any uncopied buffers.  [blitv] will never raise an exception, since
    it returns any uncopied bytes if the [dst] buffer is not big enough
    to fit the full list of [src] buffers. *)
```

via @talex5 in mirage/mirage-net-xen#17

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,6 @@
+1.6.0 (trunk):
+* Add `blitv` to copy over a list of buffers (from Thomas Leonard).
+
 1.5.0 (2014-11-24):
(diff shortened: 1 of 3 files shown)


talex5:
I already made a PR for this (but renamed the function):

https://github.com/mirage/ocaml-cstruct/pull/40

avsm:
Woops!

Issue #24, opened 2014-06-12
Equality


pqwy:
Equality on cstructs would be really useful.

Currently, one can either `Cstruct.to_bigarray` both cstructs and use bigarray equality, or walk down cstructs in OCaml. Plus, it gets reimplemented around.

A vote for a fast, canonical equality.

PR #45, opened 2015-02-02
Cstruct comparison

### LINKED ISSUE #23: fast comparison of cstructs
if we have two cstructs, which are separate parts of the memory, there should be a fast way to compare whether their contents is equal (slow way is to convert to bigarray and compare these)

pqwy:
Adds `compare` and `equal` based on `memcmp`.

Fixes #23 and #24.

### DIFF
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -181,6 +181,16 @@ val of_string: ?allocator:(int -> t) -> string -> t
     with the underlying buffer allocated by [alloc]. If [allocator] is not
     provided, [create] is used. *)
 
+(** {2 Comparison } *)
+
+val equal : t -> t -> bool
+(** [equal t1 t2] is [true] iff [t1] and [t2] correspond to the same sequence of
+    bytes. *)
+
+val compare : t -> t -> int
+(** [compare t1 t2] is [-1] if [t1] precedes [t2] by lexicographical order, [1]
+    if [t2] precedes [t1], and [0] otherwise. *)
+
 (** {2 Getters and Setters } *)
 
 val byte_to_int : byte -> int
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -188,8 +188,7 @@ val equal : t -> t -> bool
     bytes. *)
 
 val compare : t -> t -> int
-(** [compare t1 t2] is [-1] if [t1] precedes [t2] by lexicographical order, [1]
-    if [t2] precedes [t1], and [0] otherwise. *)
+(** [compare t1 t2] gives an unspecified total ordering over {!t}. *)
 
 (** {2 Getters and Setters } *)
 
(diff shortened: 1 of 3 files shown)


lib/cstruct.mli
```diff
@@ -181,6 +181,16 @@ val of_string: ?allocator:(int -> t) -> string -> t
...
 
+(** {2 Comparison } *)
+
+val equal : t -> t -> bool
+(** [equal t1 t2] is [true] iff [t1] and [t2] correspond to the same sequence of
+    bytes. *)
+
+val compare : t -> t -> int
+(** [compare t1 t2] is [-1] if [t1] precedes [t2] by lexicographical order, [1]
+    if [t2] precedes [t1], and [0] otherwise. *)
```
dbuenzli: Is the lexicographical property really useful ? If you just say that this is a total order on cstruct values you could avoid the call to C and `memcmp` whenever `t1.len <> t2.len`.

pqwy: Well it gives the order its meaning. If you postulate that `"ABC"` < `"BCD"`, but `"ABC"` > `"BC"`, it just becomes totally abstract.

OTOH the real reason for `compare` is to have a fast `equal`, `compare` just happens to pop out half the way... and I realized that `equal` should indeed short-circuit.

dbuenzli: > Well it gives the order its meaning. If you postulate that "ABC" < "BCD", but "ABC" > "BC", it just becomes totally abstract.

Yes but for example that's all that is needed to give to  `Map.Make` or `Set.Make` and since cstruct values may contain *any* kind of structures with fields that may span more than one byte I don't any value in that byte-based overspecification.

pqwy: Yeah, if it's only a consistent total order for `Map.Make` and the like, it could be made slightly faster...

I'm undecided. Third-party ideas?

yallop: The [bigarray comparison](https://github.com/ocaml/ocaml/blob/dc7e685a40c0bdcfa3373f3703bab458d9bba783/otherlibs/bigarray/bigarray_stubs.c#L704-L717) checks element types and dimensions before falling back to a lexicographic comparison, so it probably wouldn't be too surprising for cstruct to do something similar.

dbuenzli: Le lundi, 2 février 2015 à 22:28, David Kaloper a écrit :
> Yeah, if it's only a consistent total order for Map.Make and the like, it could be made slightly faster...
> I'm undecided. Third-party ideas?

Note that Pervasives polymorphic comparison is defined just like that (while it happens to sort tuples lexicographically that's actually not guaranteed by the interface spec.)

Best,

Daniel

samoht:
> I'm undecided. Third-party ideas?

lexicographic order doesn't matter, just pick a stable and efficient one.

pqwy:
OK. Convinced.

g2p:
Lexicographic order can be useful (eg for ordering sha1s like git does), but it seems that we have this if the structs share the same length.
Please keep this property in the future!

PR #48, opened 2015-02-11
Localized exn

samoht:


samoht:
would be great if someone could have a quick review on this one before I merge it.

rgrinberg:
Looks ok to me. `invalid_arg` is a bit simpler but that's about it.

samoht:
Ok rebased with comments taken into account. The new patch is bigger but it should make debugging easier.

dbuenzli:
Much better ! I will invalid arg my code only to be able to contemplate them. 

However there's one last thing. I suspect that in general the info you give will not be enough to avoid a printf debugging session. I suspect you also want to always put [this info](https://github.com/mirage/ocaml-cstruct/pull/48/files#diff-86064644d5ce33bb29237899dfa166c3R86) in the string aswell (should be easy to thread in the base `err` function).

lib/cstruct.ml
```diff
@@ -314,7 +330,10 @@ let iter lenfn pfn buf =
           body := None;
           None
         |Some plen ->
-          let p,rest = split buf plen in
+          incr i;
+          let p,rest =
+            try split buf plen with Invalid_argument _ -> err_iter !i plen
```
dbuenzli: The info of this error is not expressed in terms of the client provided arguments, or am I missing something ?

samoht: Not directly, but the `Some plen` value is returned by the `lenfn` function provided by the user (which is called multiple times on the buffer). I think it is useful to know how many time that function has been called by `iter` (the `i`) and what splitting it is trying to do (the `plen`).

samoht:
yea, having more info on the underlying buffer should be useful as well. I've almost been tempted to output an hex representation of the buffer as well, but that's maybe too much. Returning `t.off` and `t.len` is less verbose and should be useful. I'll add these.

dbuenzli:
> Returning t.off and t.len is less verbose and should be useful. I'll add these.

And the underlying bigarray's buffer length, because of `set_len`,  `add_len`, no ?

samoht:
Thanks all for the comments, merging now.

PR #49, opened 2015-02-11
Expose memset

pqwy:
This adds `memset`.

I realize adding more C to `cstruct` is not to be taken lightly, but a fast way to set all the bytes in a buffer is often desirable. `nocrypto` already uses memset on cstructs ([here](https://github.com/mirleft/ocaml-nocrypto/blob/master/src/uncommon.ml#L126) through [this](https://github.com/mirleft/ocaml-nocrypto/blob/master/src_gen/bindings.ml#L13)), at some point there was talk about initializing all constructed cstructs to `0x00`, and I imagine there might be other uses of this.

The operation is named `set` because `fill` is confusingly similar to `fillv`.

### DIFF
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -262,6 +262,9 @@ val blit_to_string: t -> int -> string -> int -> int -> unit
     valid segment of [src], or if [dstoff] and [len] do not designate
     a valid substring of [dst]. *)
 
+val set: t -> int -> unit
+(** [set t x] sets all the bytes of [t] to [x land 0xff]. *)
+
 val len: t -> int
 (** Returns the length of the current cstruct view.  Note that this
     length is potentially smaller than the actual size of the underlying
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -262,8 +262,8 @@ val blit_to_string: t -> int -> string -> int -> int -> unit
     valid segment of [src], or if [dstoff] and [len] do not designate
     a valid substring of [dst]. *)
 
-val set: t -> int -> unit
-(** [set t x] sets all the bytes of [t] to [x land 0xff]. *)
+val memset: t -> int -> unit
+(** [memset t x] sets all the bytes of [t] to [x land 0xff]. *)
 
 val len: t -> int
 (** Returns the length of the current cstruct view.  Note that this
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -118,6 +118,8 @@ external unsafe_blit_bigstring_to_string : buffer -> int -> string -> int -> int
 
 external unsafe_compare_bigstring : buffer -> int -> buffer -> int -> int -> int = "caml_compare_bigstring" "noalloc"
 
+external unsafe_fill_bigstring : buffer -> int -> int -> int -> unit = "caml_fill_bigstring" "noalloc"
+
 let copy src srcoff len =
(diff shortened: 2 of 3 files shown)


dbuenzli:
I think `set` is a confusing name, especially w.r.t. to the existing `{get,set}_*` functions of the module and the `(Big)Array.set` functions. Here are a few alternatives: `reset`, `clear`, `set_all` or even `set_all_uint8`.

lib/cstruct.mli
```diff
@@ -262,6 +262,9 @@ val blit_to_string: t -> int -> string -> int -> int -> unit
     valid segment of [src], or if [dstoff] and [len] do not designate
     a valid substring of [dst]. *)
 
+val set: t -> int -> unit
```
avsm: Shouldn't this be `t -> char -> unit` ?

avsm: Although I guess `set_uint8` is indeed an `int`, so yours is consistent.

pqwy: Yes, it's because of consistency with `set_uint8`. I think it would be confusing otherwise, as to what else is different between the two.

pqwy:
Hmm...

Travis breaks because apparently `ocamlrun` takes only `$CAML_LD_LIBRARY_PATH` into account, so when `bounds.byte` runs, only the C bits of the already installed version of `cstruct` matter.

@dbuenzli The name choice was not too fortunate. Updated to completely consistent and obvious `set_all_uint8`.

samoht:
See my other PRs for the travis fix

hannesm:
I'd expect this function here to be called `fill`.
`fillv` is slightly the wrong name for what it does, I'd be happy to rename `fillv` to `transfer` (given that there is a single release (which hasn't hit opam-repository yet) and a single use of it, renaming shouldn't hurt too much).
Talking about API, `concat : t list -> t` and `append : t -> t -> t` (from [nocrypto](https://github.com/mirleft/ocaml-nocrypto/blob/master/src/uncommon.ml#L54-L74)) would be nice to have imho - what do others think? should I PR?

avsm:
Rather than renaming fillv, I'd suggest just calling this `memset` as it's the name that will be most familiar to those coming it to use Cstruct.  I believe we didn't want to use `transfer` since it suggests mutating the source, which `fillv` doesn't (see discussion on mirage/ocaml-cstruct#40).

It must do a bounds check before it can be merged though.  We can't just directly expose unsafe functions in the Cstruct interface.

pqwy:
Force-pushed an update to rename to `memset`.

As for the bounds-check, if follows from the construction: since this summer's batch of bounds-checks, it's impossible to create a `Cstruct.t` with illegal bounds. If this holds, calling memset on the region it points to should be safe.

I can add it if you insist, but I would prefer to try to work out how to increase our confidence that this property is not violated in the first place.

avsm:
It's worth a comment in the source to note the invariant.  I wonder if we can protect that via a GADT internally to the library (secondary concern, not blocking a merge here!).

One thing I've wanted for a while is a super-paranoid `cstruct.safe` (or `overly-safe`) that just puts bounds checking everywhere when speed isn't a concern.  It would be useful to link that in instead of `cstruct` for situations where extra paranoia wins over pure performance.

pqwy:
Documented the invariant.

pqwy:
Re: `cstruct.safe` -- I wonder if we can have both at the same time...

PR #51, opened 2015-02-12
Sexp tests

pqwy:
Tests freeze the sexp format, though.


samoht:
Can you update `test.sh` instead of using `_oasis`? Or need to set-up `$CAML_LD_LIBRARY_PATH` correctly to find the shared library.

samoht:
@pqwy ping?

PR #54, opened 2015-03-28
rerelease 1.6.0

avsm:
Also fixes for sexp test compilation


samoht:
The tests need to be hooked in the Travis scripts (currently it just calls `./test`). A good fix would be to compile all the tests in `_oasis` directly.

avsm:
See #55 for test fixups

samoht:
@avsm need to call the new s-exp test in test.sh as well

avsm:
Why? Both are being run independently, and we're phasing out test.sh

samoht:
@avsm because if you look at https://github.com/mirage/ocaml-cstruct/blob/master/opam#L24 you'll see that (1) we call "configure" again which is useless and more importantly (2) we only call `./test.sh` in the tests. So either we should call `make test` there, or we should add the test into `test.sh` so it will be run by Travis.

PR #55, opened 2015-03-28
shift to centralised Travis scripts

avsm:


avsm:
Addition of `compare` breaks ocaml-dns; see mirage/ocaml-dns#62

PR #57, opened 2015-04-04
Concatenators

pqwy:
`concat` and  `append` from `nocrypto`, as [discussed](https://github.com/mirleft/ocaml-nocrypto/issues/50#issuecomment-89523111).

`concat` has a very value-like semantics in that it maps `[x]` to `x` itself without making a copy. This is beneficial in `nocrypto` as single-buffer situations can be treated as special cases of multi-buffer ones without that little bit of overhead. It's also documented, but maybe bears pointing out explicitly.

I personally like the operator form of `append`, `<+>`, but didn't include it not to extend the API too much. If others think they too prefer `foo <+> bar` to `append foo bar`, I can either add an alias or completely rename it.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -380,6 +380,24 @@ let rec fold f next acc = match next () with
   | None -> acc
   | Some v -> fold f next (f acc v)
 
+let append cs1 cs2 =
+  let l1 = len cs1 and l2 = len cs2 in
+  let cs = create (l1 + l2) in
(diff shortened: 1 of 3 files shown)


samoht:
The issue with `<+>` is that you really don't want people to write `a <+> b <+> c` (but `concat [a;b;c]`).

I'm fine with avoiding the copy when the list has only one element.

samoht:
Thanks!

PR #59, opened 2015-04-27
ppx syntax extension

nojb:
I was playing around with `ppx` and decided to port the `camlp4` extension over to `ppx`.
The only remaining issue is to improve a little bit the error messages, but wanted to get some feedback on the syntax.

This is how you declare a `cstruct`:
```ocaml
[%%cstruct
type ethernet = {
  dst : uint8_t [@l 6];
  src : uint8_t [@l 6];
  ethertype : uint16_t
} [@@big_endian]
]
```
(the `[@l n]` attribute means an array of *length* `n`).

And a `cenum`:
```ocaml
[%%cenum
type bar16 =
  | ONE [@id 1]
  | TWO
  | FOUR [@id 4]
  | FIVE
  [@@uint16_t] [@@sexp]
]
```

Thoughts ? Comments ?


nojb:
Also related, see https://github.com/nojb/ppx_bitstring.  I ported [`ocaml-bitstring`](https://code.google.com/p/bitstring/) to `ppx` and am working on adding `bigarray` (and hence `cstruct`) support.

avsm:
Lovely!  One first gut reaction is how much harder to read the new syntax is to read than the cleaner `camlp4` one.  Is it necessary to break down the grammar within the `[%%cstruct ...]` to use attributes for the various decorators (like id, sexp, or the number format).

It would just be easier and more readable to port code by wrapping it in the `[%%cstruct ]` and leaving the internal grammar unchanged. However, I can see why it's beneficial to be more fine-grained for the purposes of IDE support and other tools manipulating this AST in the future.

avsm:
The local `opam` file needs updating for Travis to pick up the ppx:

```
- ocamlfind: Package `ppx_tools.metaquot' not found
- W: Field 'pkg_ppx_tools_metaquot' is not set: Command ''/home/travis/.opam/system/bin/ocamlfind' query -format %d ppx_tools.metaquot > '/tmp/oasis-cca56b.txt'' terminated with error code 2
- E: Cannot find findlib package ppx_tools.metaquot
```

nojb:
Updated `opam`.  I am not sure how to convince `oasis` to generate the right `ppx` entries in the `META` file.

Regarding the syntax, yes, indeed, we could preserve the existing syntax with the quoting mechanism `{|...|}`, with the tradeoff that you mention- you lose all AST manipulating ability.  Nonetheless I will give it shot to see how it goes.

dbuenzli:
Le lundi, 27 avril 2015 à 03:23, Nicolas Ojeda Bar a écrit :
> the [@l n] attribute means an array of length n

I would use @len or @length.  

Best,

Daniel

nojb:
I experimented using `{|...|}` to preserve the existing syntax, but am not very happy with the result.  I don't think I like the tradeoff between readability and complexity very much (e.g. one needs to implement an ad hoc lexer & parser for the stuff inside `{| ... |}`).

Here are all the possibilities I have been thinking about (including one using `bitstring-ppx` just for comparison sake):

- Original one (not quite achievable using `ppx`):
```ocaml
cstruct ethernet {
  uint8_t[6] dst;
  uint8_t[6] src;
  uint16_t ethertype
} as big_endian
```
- Using `{| ... |}`:
```ocaml
let%cstruct ethernet = {|
  uint8_t[6] dst;
  uint8_t[6] src;
  uint16_t ethertype
|} [@big_endian]
```
or
```ocaml
let%cstruct ethernet [@big_endian] = {|
  uint8_t[6] dst;
  uint8_t[6] src;
  uint16_t ethertype
|}
```
- Using attributes:
```ocaml
[%%cstruct
type ethernet = {
  dst : uint8_t [@len 6];
  src : uint8_t [@len 6];
  ethertype : uint16_t
} [@@big_endian]
]
```
One source of uglyness is that we need to wrap the `cstruct` declaration with `[%%cstruct ...]`.  It would improve things if we could use the following shorthand:
```ocaml
type%cstruct = ...
```
instead of
```ocaml
[%%cstruct type = ...]
```
This is not currently supported by the compiler but it seems a reasonable suggestion in light of the shorthand notation for extension of expression terms (like `let%foo`, etc.).
- For comparison, a possible version using `bitstring-ppx`:
```ocaml
match%bitstring buf with
| dst [@len 6 * 8] [@string],
  src [@len 6 * 8] [@string],
  ethertype [@len 16] -> ...
| _ -> ...
```
What do you think ? Any suggestions/preferences/ideas ?

dbuenzli:
> Any suggestions/preferences/ideas ?

I think the attribute version is best as, except for the attribute names and payloads, it doesn't require me to learn new things from a syntactic point of view, you remain completely in the OCaml language definition (and I personally don't find it "much harder" to read).

samoht:
does the following would work?

```
type ethernet = {
  dst : uint8_t [@len 6];
  src : uint8_t [@len 6];
  ethertype : uint16_t
} [@@cstruct @@big_endian]
```

nojb:
One can attach any number of attributes to the declaration, like:
```ocaml
type ethernet = {
  dst : uint8_t [@len 6];
  src : uint8_t [@len 6];
  ethertype : uint16_t
} [@@cstruct] [@@big_endian] [@@sexp] ...
```
The problem is that in general the end result will depend on the order of application of `ppx` processors.

For example, note that in the case of `cstruct` there is no type declaration after it is done with it.

samoht:
I think the cstruct extension should generate a corresponding type then -- and maybe a general `to_cstruct` and `of_cstruct` instead of just field converters....

dbuenzli:
> For example, note that in the case of cstruct there is no type declaration after it is done with it.

Then I think the `cstruct` should not be an attribute on the type, it would be misleading.

nojb:
Agree.  Regarding @samoht's suggestion that `cstruct` should generate a corresponding type: in that case one needs to specify the type/kind of the individual fields; for example,
```ocaml
type ethernet = {
  dst : string [@len 6];
  src : string [@len 6];
  ethertype : int [@len 2] [@unsigned]
} [@@cstruct] [@@big_endian] [@@sexp]
```
The problem of compatibility with other `ppx` extension is not completely gone though: those other extensions need to be recognise attributes such as `[@unsigned]` in order to play nicely with this one.

dbuenzli:
> The problem of compatibility with other ppx extension is not completely gone though:

I'd say forget about this. `ppx`s are inherently broken from a compositional point of view. You may get to something, but it will so brittle I'm not sure it's worth having.

nojb:
There is another possibility mentioned in the past which is that this syntax extension really belongs in `ctypes` as a way to specify C structure types.  This would stay closer to the current state of affairs in that no underlying type would be generated.  For example, the declaration
```ocaml
[%%cstruct type ethernet = {
  dst : uint8_t [@len 6];
  src : uint8_t [@len 6];
  ethernet : uint16_t
} [@@big_endian] ]
```
could generate the following code (to be considered in the scope of `Ctypes`):
```ocaml
type ethernet
let ethernet : ethernet structure typ = structure "ethernet"
let ethernet_dst = field ethernet "dst" (array 6 uint8_t)
let ethernet_src = field ethernet "src" (array 6 uint8_t)
let ethernet_ethertype = field ethernet "ethertype" uint16_t
let () = seal ethernet
let ethernet_of_cstruct cs : ethernet structure =
  let p = bigarray_start array1 (Cstruct.to_bigarray cs) cs in
  @! (coerce (ptr char) (ptr ethernet) p)
let get_ethernet_dst e : char carray =
  getf e ethernet_dst
...
```
(endianness functions should be added to `Ctypes` if necessary)

I actually like this idea quite a bit because this would be of use more generally as well.

What do you think ?

yallop:
> There is another possibility mentioned in the past which is that this syntax extension really belongs in ctypes as a way to specify C structure types. [...] For example, the declaration
> 
> ```
> [%%cstruct type ethernet = {
>   dst : uint8_t [@len 6];
>   src : uint8_t [@len 6];
>   ethernet : uint16_t
> } [@@big_endian] ]
> ```
> could generate the following code [...]

Avoiding duplicating the layout calculations in cstruct and ctypes seems like a good idea.  If you're using the public API then the syntax extension doesn't necessarily have to live inside the main ctypes distribution, AFIACS.

What do you think about using C syntax rather than OCaml syntax?

```C
[%%C struct ethernet {
  uint8_t dst[6];
  uint8_t src[6];
  uint16_t ethernet;
} [@@big_endian] ]
```

It seems that C syntax might lend itself more naturally to extensions, e.g.:

```C
[%%C struct sockaddr {
  unsigned short sa_family;
  union {
    struct sockaddr_in in;
    struct sockaddr_in6 in6;
    /* ... */
  };
} [@@big_endian] ]
```

> (endianness functions should be added to Ctypes if necessary)

We have an issue open for this:

  https://github.com/ocamllabs/ocaml-ctypes/issues/84

avsm:
C syntax would certainly be easier to cut-and-paste from external header files (and eventually, perhaps even auto-import)

Drup:
Ping ?

(Note: @yallop's "C syntax" is not parsable by OCaml, @nojb's previous proposal using `{| ... |}` is the only way to go if you want C syntax)

Drup:
rebased with some fixes in https://github.com/Drup/ocaml-cstruct/tree/ppx

samoht:
@Drup can you create a PR with your changes?

avsm:
Apologies for dropping this PR; @drup a new PR would be appreciated and I will review.

nojb:
I'm closing this to shift focus to @Drup's (upcoming) PR.

avsm:
ping @drup -- any chance of a PR of your changes?

PR #64, opened 2015-05-31
Add javascript stubs

### LINKED ISSUE #63: A .js subpackage?
I was just experimenting creating a 'tutorial' iocamljs notebook for [mirage/ocaml-vchan] in the style of [avsm/ocaml-dockerfile]. Several Mirage libraries use a small number of C bindings which would need to be implemented somewhere for javascript.

Does it make sense to create a `cstruct.js` which would contain the needed external functions?

djs55:
This allows javascript programs to use the full cstruct library.

Note: to link the stubs you must add the argument
  `+cstruct/cstruct.js`
to the js_of_ocaml command-line.

Fixes #63

Signed-off-by: David Scott <dave.scott@citrix.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,11 @@ configure:
 
 # OASIS_STOP
 
+JS_DIR ?= $(shell ocamlfind query cstruct)
+
 .PHONY: js-install js-uninstall
 js-install:
-	install -m 0644 js/cstruct.js $(shell ocamlfind query cstruct)
+	install -m 0644 js/cstruct.js $(JS_DIR)
 
 js-uninstall:
-	rm $(shell ocamlfind query cstruct)/cstruct.js
+	rm $(JS_DIR)/cstruct.js
--- a/Makefile
+++ b/Makefile
@@ -39,3 +39,10 @@ configure:
 .PHONY: build doc test all install uninstall reinstall clean distclean configure
 
 # OASIS_STOP
+
+.PHONY: js-install js-uninstall
+js-install:
+	install -m 0644 js/cstruct.js $(shell ocamlfind query cstruct)
+
+js-uninstall:
+	rm $(shell ocamlfind query cstruct)/cstruct.js
--- a/opam
+++ b/opam
@@ -32,8 +32,14 @@ build-test: [
   [make]
   ["./test.sh"]
 ]
-install: [make "install"]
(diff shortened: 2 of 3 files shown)


djs55:
One of the builds failed with an error in an unrelated package:
```
### stderr ###
# test.c:1:17: fatal error: gmp.h: No such file or directory
# compilation terminated.
```

Makefile
```diff
@@ -39,3 +39,10 @@ configure:
 .PHONY: build doc test all install uninstall reinstall clean distclean configure
 
 # OASIS_STOP
+
+.PHONY: js-install js-uninstall
+js-install:
+	install -m 0644 js/cstruct.js $(shell ocamlfind query cstruct)
```
avsm: could this shell invocation be stuck into a Make variable instead so it can be manually overridden?  I do worry about cross compilation these days.

opam
```diff
@@ -32,8 +32,14 @@ build-test: [
   [make]
   ["./test.sh"]
 ]
-install: [make "install"]
-remove:  ["ocamlfind" "remove" "cstruct"]
+install: [
+  [make "install"]
+  [make "js-install"]
```
avsm: should this be depopted on `js_of_ocaml`, or is it not worth the bother (just install it unconditionally)

samoht:
I've restarted the build.If we move these stubs upstream (as you already did) do we really need this PR?

djs55:
These stubs are cstruct specific-- the jsofocaml ones are for the extra
intrinsics added to ocaml to support ocplib-endian iirc

I think everywhere we've got some C fns, we ought to have js versions too...

Makefile
djs55: Good idea

opam
djs55: I thought it wasn't worth the bother: nothing needs to be built, so it seems artificial to rebuild cstruct (and all packages dependent on it) if we install js_of_ocaml.

Ideally in future ocamlfind and oasis would know about js stubs as well as C stubs and this would all be automatic.

samoht:
Can you rebase on master? you need some fixes on `.travis.yml` to make the tests pass. Thanks!

djs55:
Will do!

samoht:
Thanks!

PR #69, opened 2015-07-11
Remove name/version in opam file

samoht:
/cc @avsm these fields are usually not needed and it's confusing to keep them in sync with the rest of the metadata.

### DIFF
--- a/opam
+++ b/opam
@@ -1,6 +1,4 @@
 opam-version: "1.2"
-name:         "cstruct"
-version:      "1.6.0"
 maintainer:   "anil@recoil.org"
 authors:      "Anil Madhavapeddy"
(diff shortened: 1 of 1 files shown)


dsheets:
It's nice for pinning, though, and it means you can write `opam publish prepare [URL]` instead of `opam publish prepare cstruct.1.6.0 [URL]`. But yes, it's unfortunate that there isn't a better way to keep them up-to-date in a dev repo.

avsm:
Indeed, they're needed for pinning...

samoht:
Not really, you can add the version on the command-line if you really need the version:

```
opam pin add cstruct.42 --dev
```

PR #74, opened 2015-10-20
Support for -safe-string

yallop:
Add support for compiling with `-safe-string`.  The main change is to rename `blit_to_string` to `blit_to_bytes` and change its type so that it writes to `bytes` rather than `string`:

```ocaml
val blit_to_bytes: t -> int -> Bytes.t -> int -> int -> unit
```

The existing name is kept as an alias, but marked as deprecated.


yallop:
This is now ready to merge.

PR #75, opened 2015-11-04
correct error output for Cstruct.LE.get_uint16 on invalid bounds

yomimono:
fix an incorrect string used in error reporting when bounds check fails in Cstruct.LE.get_uint16 .


avsm:
thanks!

PR #76, opened 2015-11-05
PPX support, take 2 

avsm:

PR #81, opened 2016-02-14
Fix %%cenum in ppx

hannesm:
as proposed by @nojb in https://github.com/mirage/ocaml-cstruct/pull/76#issuecomment-183765162

This works for me.

### DIFF
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -320,7 +320,7 @@ let output_enum _loc name fields width ~sexp =
   let printers = List.map (fun ({txt = f},_) ->
       {pc_lhs = Ast.pconstr f []; pc_guard = None; pc_rhs = Ast.str f}) fields in
   let parsers = List.map (fun ({txt = f},_) ->
-      {pc_lhs = Ast.pstr f; pc_guard = None; pc_rhs = Ast.constr f []}) fields in
+      {pc_lhs = Ast.pstr f; pc_guard = None; pc_rhs = Ast.constr "Some" [Ast.constr f []]}) fields in
   let getter {txt = x} = sprintf "int_to_%s" x in
   let setter {txt = x} = sprintf "%s_to_int" x in
   let printer {txt = x} = sprintf "%s_to_string" x in
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -252,7 +252,7 @@ let output_hexdump _loc s =
                 [%expr Printf.bprintf _buf "0x%Lx\n" ([%e Ast.evar (getter_name s f)] v)]
               |Buffer (_,_) ->
                 [%expr Printf.bprintf _buf "<buffer %s>"
-                         [%e Ast.str (field_to_string f)]
+                         [%e Ast.str (field_to_string f)];
                          Cstruct.hexdump_to_buffer _buf ([%e Ast.evar (getter_name s f)] v)]
           ]]
       ) (Ast.unit ()) s.fields
@@ -407,13 +407,19 @@ let constr_enum = function
   | {pcd_loc = loc} ->
     loc_err loc "invalid cenum variant"
 
-let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc} =
-  let sz = match fty.ptyp_attributes with
+let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc; pld_attributes = att} =
+  let get = function
     | [{txt = "len"}, PStr
          [{pstr_desc = Pstr_eval ({pexp_desc = Pexp_constant (Const_int sz)}, _)}]] ->
       Some sz
     | _ ->
       None
+  in
+  let sz = match get fty.ptyp_attributes, get att with
+  | Some sz, None
+  | None, Some sz -> Some sz
+  | Some _, Some _ -> loc_err loc "multiple field length attribute"
+  | None, None -> None
   in
   let fty = match fty.ptyp_desc with
     | Ptyp_constr ({txt = Lident fty}, []) -> fty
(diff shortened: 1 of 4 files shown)


hannesm:
ping, any takers? @avsm? the most recent release does support ppx, but `cenum` is broken... this PR, code from @nojb, fixes cenum... would be great to have a minor fix release with this in, so that several of my libraries can switch over to ppx...

avsm:
Looks good to me -- I got blocked on adding a test suite.  Any chance you could push a file with `lib_test/basic.ml` into `ppx_test/basic.ml` with an example of the cstruct and cenum syntax? I'll fix up the build...

hannesm:
I did so (added basic.ml and enum.ml in ppx), and they fail:
```
File "ppx_test/basic.ml", line 75, characters 2-11:
Error: This function has type Cstruct.t -> Cstruct.uint8 -> unit
       It is applied to too many arguments; maybe you forgot a `;'.
```

The offending line is: `set_foo_d s1 0 be;`, where `d` in `foo` is a `uint8_t [@len 8]`.

Seems cstruct.ppx is not completely drop-in for the camlp4 counterpart... any ideas/quick fixes @nojb @Drup?

hannesm:
NB someone should normalise the build system here, using both `_oasis` and a custom `test.sh` is very confusing...

hannesm:
If I write instead `d : (uint_8 [@len 8])`, I get the following error:
```
File "_none_", line 1:
Error: This expression has type
         ('a -> 'b -> 'c -> 'd, Buffer.t, unit, unit, unit,
          'a -> 'b -> 'c -> 'd)
         CamlinternalFormatBasics.fmt
       but an expression was expected of type
         ('a -> 'b -> 'c -> 'd, Buffer.t, unit, unit, unit, unit)
         CamlinternalFormatBasics.fmt
       Type 'a -> 'b -> 'c -> 'd is not compatible with type unit 
```

nojb:
Hi @hannesm, the error is related to the fact that [@len] attribute is being attached to the *field* rather than the type of the field (the way this worked changed a few months ago).  The following patch should fix this.  Thanks!

```diff
diff --git a/ppx/ppx_cstruct.ml b/ppx/ppx_cstruct.ml
index e6a781f..043f8cf 100644
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -407,14 +407,20 @@ let constr_enum = function
   | {pcd_loc = loc} ->
     loc_err loc "invalid cenum variant"
 
-let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc} =
-  let sz = match fty.ptyp_attributes with
+let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc; pld_attributes = att} =
+  let get = function
     | [{txt = "len"}, PStr
          [{pstr_desc = Pstr_eval ({pexp_desc = Pexp_constant (Const_int sz)}, _)}]] ->
       Some sz
     | _ ->
       None
   in
+  let sz = match get fty.ptyp_attributes, get att with
+  | Some sz, None
+  | None, Some sz -> Some sz
+  | Some _, Some _ -> loc_err loc "multiple field length attribute"
+  | None, None -> None
+  in
   let fty = match fty.ptyp_desc with
     | Ptyp_constr ({txt = Lident fty}, []) -> fty
     | _ ->
```

nojb:
NB The patch is to solve the earlier problem, please check if you still get the error involving format strings after applying it.

nojb:
OK, I think the second error is due to a typo.  Patch incoming.

nojb:
It was a missing semicolon.  The following patch should fix both problems:

```diff
diff --git a/ppx/ppx_cstruct.ml b/ppx/ppx_cstruct.ml
index e6a781f..533c1a8 100644
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -252,8 +252,8 @@ let output_hexdump _loc s =
                 [%expr Printf.bprintf _buf "0x%Lx\n" ([%e Ast.evar (getter_name s f)] v)]
               |Buffer (_,_) ->
                 [%expr Printf.bprintf _buf "<buffer %s>"
-                         [%e Ast.str (field_to_string f)]
-                         Cstruct.hexdump_to_buffer _buf ([%e Ast.evar (getter_name s f)] v)]
+                         [%e Ast.str (field_to_string f)];
+                       Cstruct.hexdump_to_buffer _buf ([%e Ast.evar (getter_name s f)] v)]
           ]]
       ) (Ast.unit ()) s.fields
   in
@@ -407,14 +407,20 @@ let constr_enum = function
   | {pcd_loc = loc} ->
     loc_err loc "invalid cenum variant"
 
-let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc} =
-  let sz = match fty.ptyp_attributes with
+let constr_field {pld_name = fname; pld_type = fty; pld_loc = loc; pld_attributes = att} =
+  let get = function
     | [{txt = "len"}, PStr
          [{pstr_desc = Pstr_eval ({pexp_desc = Pexp_constant (Const_int sz)}, _)}]] ->
       Some sz
     | _ ->
       None
   in
+  let sz = match get fty.ptyp_attributes, get att with
+  | Some sz, None
+  | None, Some sz -> Some sz
+  | Some _, Some _ -> loc_err loc "multiple field length attribute"
+  | None, None -> None
+  in
   let fty = match fty.ptyp_desc with
     | Ptyp_constr ({txt = Lident fty}, []) -> fty
     | _ ->
```

hannesm:
thx @nojb, I applied those in this PR. the build system seems to not work without cstruct already installed (oasis configure complains about missing package cstruct.ppx) -- needs some yak shaving in test.sh (and move the ppx tests into there)... @avsm I hope you will fix that build system!?

hannesm:
(re-removed the oasis bits, and got it into test.sh, works at least locally (travis thingy seems to be broken here..))

Issue #84, opened 2016-02-27
unnecessary hard type_conv dependency


hannesm:
type_conv pulls in camlp4, but there might be individuals who want to build cstruct without camlp4 support -- i'm not sure what the correct formulation in opam is, but using sth like ` "--%{camlp4+type_conv:enable}%-camlp4"`, plus moving type_conv to depopts should do the trick!?!

hannesm:
fixed in #95

PR #85, opened 2016-02-27
don't depend hard on type_conv

hannesm:
attempt at fixing #84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ script: bash -ex ./.travis-docker.sh
 env:
   global:
   - PACKAGE="cstruct"
-  - DEPOPTS="camlp4 ppx_tools async lwt"
+  - DEPOPTS="camlp4 type_conv ppx_tools async lwt"
   - REVDEPS="tls dns irmin tcpip"
   matrix:
-  - DISTRO="debian-stable" OCAML_VERSION="4.01.0" DEPOPTS="camlp4 lwt"
+  - DISTRO="debian-stable" OCAML_VERSION="4.01.0" DEPOPTS="camlp4 type_conv lwt"
   - DISTRO="debian-testing" OCAML_VERSION="4.02.3"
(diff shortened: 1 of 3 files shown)


hannesm:
what now fails are the reverse dependencies, who expect that depending on cstruct will give them cstruct.syntax -- whereas they need to depend on type_conv + camlp4 (actually, type_conv is sufficient since it depends on camlp4).  any opinions? should we fix up all the cstruct.syntax users? @avsm @samoht others?

samoht:
I didn't know that we had `cstruct.syntax` user actually :-)
A possibility is to add a virtual `cstruct-syntax` package and use this explicitly in the constraints instead of some magic `camlp4+type_conv` incantations.

samoht:
or actually, just adding version constraints to existing users of `cstruct.syntax` should be enough...

hannesm:
@samoht well, tcpip, dns, ... all use `cenum` and `cstruct` p4 extension. adding a `cstruct-syntax` package sounds like a good idea

hannesm:
a version constraint would not be a good solution imho, since cstruct keeps to provide that sub-package if you have type_conv and camlp4 installed...

avsm:
> a version constraint would not be a good solution imho, since cstruct keeps to provide that sub-package if you have type_conv and camlp4 installed...

hm yeah, we want to propagate additional dependencies...

avsm:
> A possibility is to add a virtual cstruct-syntax package and use this explicitly in the constraints instead of some magic camlp4+type_conv incantations.

Yes, I prefer this solution... reducing the depopts (and especially combinations of them) is the way to go...

hannesm:
but virtual packages feel very hackish as well.. esp since you might need to version bump them at arbitrary points (and people tend to forget this during release processes - see mirage-types/mirage-types-lwt in opam-repo)... I'd actually use the type_conv instead of camlp4 depopt, since type_conv depends on camlp4 anyways (or just stick with both camlp4 and type_conv)

hannesm:
superseeded by #95 I believe

PR #87, opened 2016-03-06
Add function to check buffer alignment

djs55:
Some protocols require buffers to be aligned to particular boundaries, for example

- Xen shared memory pages must be page-aligned
- Linux O_DIRECT I/O buffers must be sector-aligned

This PR adds a simple `check_alignment` function (similar to `check_bounds`) which returns `true` if the data inside a struct (i.e. the first byte at the offset) has an address which is suitably aligned.

This will enable better sanity-checking and error-reporting in other Mirage libraries.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -95,11 +95,9 @@ let create len =
 let check_bounds t len =
   Bigarray.Array1.dim t.buffer >= len
 
-external address_bigstring : buffer -> int64 = "caml_address_bigstring"
+external check_alignment_bigstring : buffer -> int -> int -> bool = "caml_check_alignment_bigstring"
 
-let check_alignment t alignment =
-  let address = address_bigstring t.buffer in
-  Int64.(rem (add address (of_int t.off)) (of_int alignment) = 0L)
+let check_alignment t alignment = check_alignment_bigstring t.buffer t.off alignment
 
 type byte = char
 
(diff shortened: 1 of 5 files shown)


lib/cstruct.ml
```diff
@@ -95,6 +95,12 @@ let create len =
 let check_bounds t len =
   Bigarray.Array1.dim t.buffer >= len
 
+external address_bigstring : buffer -> int64 = "caml_address_bigstring"
+
+let check_alignment t alignment =
+  let address = address_bigstring t.buffer in
+  Int64.(rem (add address (of_int t.off)) (of_int alignment) = 0L)
```
avsm: There are an awful lot of minor heap allocations here... if this is in the fast path as an assert it will probably increase GC pressure

lib/cstruct_stubs.c
```diff
@@ -69,3 +69,10 @@ caml_fill_bigstring(value val_buf, value val_ofs, value val_len, value val_byte)
          Long_val(val_len));
   return Val_unit;
 }
+
+CAMLprim value
+caml_address_bigstring(value val_buf)
+{
+  uint64_t address = (uint64_t) Caml_ba_data_val(val_buf);
+  return caml_copy_int64(address);
```
avsm: I think it might be better to check the alignment of the pointer directly in C...?

djs55:
Yeah good point, I'll respin with a C version.

PR #88, opened 2016-03-07
 PPX package version wasn't bumped

orbifx:


orbifx:
Also, the description is only mentioning `camlp4` ("access C structures via a camlp4 extension") and not `ppx`.

djs55:
Ah yes, good catch!

PR #89, opened 2016-03-12
Fix tests on windows and add appveyor

djs55:
The test.sh was failing only because it used symlinks; switch to `cp` instead.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,13 @@
+platform:
+  - x86
+
+environment:
+  CYG_ROOT: "C:\\cygwin"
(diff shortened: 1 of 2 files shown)


djs55:
I don't think appveyor actually ran the tests, but I ran them locally and they're working fine for me.

samoht:
Can you enable the tests on appveyor? It's a matter of adding a `-t` here: https://github.com/ocaml/ocaml-ci-scripts/blob/master/appveyor-opam.sh#L46

djs55:
Will do

djs55:
see [ocaml/ocaml-ci-scripts#92]

PR #90, opened 2016-03-12
Add LICENSE file (ISC)

### LINKED ISSUE #61: License file is missing
While the source indicates that this is MIT licensed, without the license file its hard to tell at a glance. For companies with a legal department it could also be an issue of which files are open and which are not. If you add a license file that just removes all the confusion.

djs55:
The Copyright (c) lines have been prepopulated from the commit
Author fields.

Fixes #61

Signed-off-by: David Scott <dave.scott@docker.com>


dsheets:
Seems to miss me. I don't particularly care but something may be broken in your automation (unless I misunderstand what dataset you are processing).

djs55:
Oer! Sorry, I'll reprocess!

PR #100, opened 2016-05-04
Add hexdump_pp

talex5:
This works better with the Logs library than using hexdump_to_buffer.
It also makes it easy to indent the hexdump.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -336,23 +336,19 @@ let of_string ?allocator buf =
     blit_from_string buf 0 c 0 buflen;
     set_len c buflen
 
-let hexdump t =
+let hexdump_pp fmt t =
   let c = ref 0 in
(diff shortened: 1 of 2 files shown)


samoht:
Why not using https://github.com/mirage/ocaml-hex/blob/master/lib/hex.mli#L60 ?

talex5:
@samoht It doesn't look like that library uses a formatter either. [hexdump_pp] is just the normal Cstruct hexdump, but for use with Format. Changing cstruct's output style might be good, but I think it's a separate issue.

talex5:
Actually, we could now rewrite the other hexdump functions in terms of `hexdump_pp`...

talex5:
I've updated the code to use the same printing logic for all three cases. The code is now shorter with the extra function than without ;-) This should also make it easier to change the logic in future, if desired.

PR #102, opened 2016-05-14
Make create zero out the new buffer

### LINKED ISSUE #30: create doesn't zero the struct
The structure returned by create isn't initialised. In a Mirage service, it might contain confidential data (e.g. TLS key material), which could easily be leaked.

Possibly this should be fixed in OCaml's Bigarray (and String.create), but at the very least, the documentation should mention this.

talex5:
The API should be safe by default. The new `create_unsafe` function can be used if you want to trade safety for speed.

Fixes #30

See also: https://github.com/mirage/mirage-tcpip/pull/204

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -88,7 +88,7 @@ let of_bigarray ?(off=0) ?len buffer =
 let to_bigarray buffer =
   Bigarray.Array1.sub buffer.buffer buffer.off buffer.len
 
-let create len =
+let create_unsafe len =
   let buffer = Bigarray.(Array1.create char c_layout len) in
   { buffer ; len ; off = 0 }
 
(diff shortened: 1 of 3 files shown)


lib/cstruct.ml
```diff
@@ -88,10 +88,15 @@ let of_bigarray ?(off=0) ?len buffer =
...
   Bigarray.Array1.sub buffer.buffer buffer.off buffer.len
 
-let create len =
+let create_unsafe len =
   let buffer = Bigarray.(Array1.create char c_layout len) in
   { buffer ; len ; off = 0 }
 
+let create len =
+  let t = create_unsafe len in
+  Bigarray.Array1.fill t.buffer (Char.chr 0);
```
hannesm: shouldn't we reuse the `memset` we define ~110 lines below?

talex5: Is there any benefit? I guess `memset` is only needed if you want to fill a sub-region of an array (`fill` can only fill the whole thing, but that's what we want here).

hannesm: consistency is my argument (I've to keep less symbols in my head when reading the code)

talex5:
OK, it uses `memset` now.

hannesm:
imho good to go (apart from a missing changes entry)! thanks!

yomimono:
I'm yomimono and I approve this PR.

PR #105, opened 2016-07-13
Add blit_from_bytes and of_bytes functions

### LINKED ISSUE #103: of_bytes?
What is the best way to take a `Bytes.t` and get it into a `Cstruct.t`?

The best I know of right now is `Cstruct.of_string (Bytes.to_string bytes)` which copies unnecessarily and is quite clumsy. Is there a better way? Should there be an `of_bytes` in the interface?

yallop:
In common with the existing implementation, this relies on the representation of `bytes` and `string` being the same.

Fixes #103.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -65,6 +65,12 @@ let err_blit_from_string_src src dst =
 let err_blit_from_string_dst src dst =
   err "Cstruct.blit_from_string src=%a dst=%a dst-off=%d len=%d"
     string_t src pp_t dst
+let err_blit_from_bytes_src src dst =
+  err "Cstruct.blit_from_bytes src=%a dst=%a src-off=%d len=%d"
+    bytes_t src pp_t dst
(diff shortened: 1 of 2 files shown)


avsm:
Can we put in some safety check about the representation change, so we don't break in the future?  I don't really understand how the safe-string evolution is meant to be happening...

yallop:
I'm not sure anyone knows yet, but there's a PR open for a [compiler-configure-time switch](https://github.com/ocaml/ocaml/pull/687) that essentially switches `-safe-string` on always and everywhere.

I think that it's quite likely that if the representation of `bytes` or `string` changes then `Bytes.unsafe_to_string` will also change, whether by being made safe, or deprecated, either of which would be fine.

In the short term the real problems are likely to crop up with code that makes assumptions about sharing and mutability, which is not the case here.

yomimono:
I'm inclined to merge this, and will do so soon unless there are objections.

PR #107, opened 2016-08-01
Add Cstruct.subv

Chris00:
The function `val subv : t list -> int -> int -> t list` does for lists of `Cstruct.t` what `sub` does for a single one.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -74,6 +74,7 @@ let err_blit_to_bytes_dst src dst=
 let err_invalid_bounds f =
   err "invalid bounds in Cstruct.%s %a off=%d len=%d" f pp_t
 let err_split = err "Cstruct.split %a start=%d off=%d" pp_t
+let err_copyv = err "Cstruct.copyv off=%d len=%d"
 let err_iter = err "Cstruct.iter %a i=%d len=%d" pp_t
 
 let of_bigarray ?(off=0) ?len buffer =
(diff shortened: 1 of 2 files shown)


yomimono:
I think this needs a name other than `subv`.  I'd expected it to more closely approximate the behavior of `List.map` with `Cstruct.sub` as an argument, but it's doing something smarter than that and probably deserves a name that reflects that.

Chris00:
@yomimono I am open to suggestions.  I picked `subv` because [lenv](https://github.com/mirage/ocaml-cstruct/blob/master/lib/cstruct.mli#L436) and [copyv](https://github.com/mirage/ocaml-cstruct/blob/master/lib/cstruct.mli#L439) (sic) do what [len](https://github.com/mirage/ocaml-cstruct/blob/master/lib/cstruct.mli#L315) and [to_string](https://github.com/mirage/ocaml-cstruct/blob/master/lib/cstruct.mli#L336) but for a sequence of Cstruct instead of a single one.

Maybe a submodule `V` with these operations (`V.len`, `V.to_string`, `V.sub`,...) is better?  (Would also allow to rename `copyv` without breaking past code.)

XVilka:
By the way, have you decided anything on this? Not that I need this feature, it is just sad to see stalled PR.

dinosaure:
Let's talk about a better name and we will able to merge it then.

avsm:
I like the suggestion of a submodule `V` or `L` (for List) with all the `*v` functions in there.  That could go in Cstruct_cap as part of the new API.

hannesm:
I haven't looked into the implementation, but what is the use case for this function?

Chris00:
@hannesm Can't remember what was the motivation at that time, sorry.

talex5:
Closing this until someone remembers what it's for (please re-open if you find a use for it).

PR #111, opened 2016-08-16
ppx_cstruct 4.04

### LINKED ISSUE #110: ppx_cstruct seems broken on 4.04
Two independent `ppx_struct` users, arp and mirage-profile, are broken on 4.04+beta on what seems to be a ppx_cstruct issue:

<http://opam.ocamlpro.com/builder/html/mirage-profile/mirage-profile.0.7.0/d5a819fca03c2e5d0f175590b6022d12>

```
+ /home/lefessan/builder/4.04.0+beta1/.opam/4.04.0+beta1/bin/ocamlfind ocamldep -package cstruct -package cstruct.ppx -package lwt -package ocplib-endian.bigstring -modules lib/trace.mli > lib/trace.mli.depends
File "lib/trace.mli", line 1:
Error: External preprocessor does not produce a valid file
Command line: ppx_cstruct '/tmp/camlppx916bcf' '/tmp/camlppx918db5'
```

<http://opam.ocamlpro.com/builder/html/arp/arp.0.1.1/f4ae306471f10b5f44ffa3090684c677>

```
+ ocamlfind ocamldep -package cstruct -package cstruct.ppx -package result -modules src/arp_wire.ml > src/arp_wire.ml.depends
File "src/arp_wire.ml", line 1:
Error: External preprocessor does not produce a valid file
Command line: ppx_cstruct '/tmp/camlppxeb84fe' '/tmp/camlppx986d5f'
```

gasche:
This PR implements support for 4.04 by just copying the 4.03 implementation which seems to work unchanged -- the local testsuite passes. This is an attempt to fix #110.

The first two commit make small changes on the `select_ppx_version` hack (see commit messages):

- the first uses an invalid source file to make sure there is a compilation failure when the OCaml version is unknown
- the second makes sure the auto-generated file is cleaned on `distclean`.

It would be nice if a release with this PR or a similar change to bring 4.04 could be made before the actual 4.04 release, so that we can test (and possibly fix) other software depending on `cstruct` during the beta/release-candidate period.


select_ppx_version.sh
```diff
@@ -4,7 +4,7 @@ V=`ocamlc -version`
 case $V in
 4.02*) SRC=402 ;;
 4.03*) SRC=403 ;;
-*) echo Unsupported OCaml version $V ;;
+*) SRC=error; echo Unsupported OCaml version $V ;;
```
hannesm: isn't an `exit 1` sufficient, rather than distributing invalid ml code for the error case?

gasche: I'll try.

gasche: Indeed it works fine, I'll rebase the PR. Thanks @hannesm!

select_ppx_version.sh
```diff
@@ -1,10 +1,17 @@
...
+  conf) (* continue *) ;;
+  *) echo "supported actions: conf, clean"; exit 1 ;;
+esac
+
 V=`ocamlc -version`
 case $V in
 4.02*) SRC=402 ;;
 4.03*) SRC=403 ;;
-*) echo Unsupported OCaml version $V ;;
+4.04*) SRC=404 ;;
```
hannesm: instead of duplicating code in the repo, we could set `SRC=403` here.. I'm still not sure which is less worse - pest or cholera

gasche: I considered it, but I think that it's better to be clear (and dumb) rather than try to be cunning inside a general approach that is rather duplication-friendly anyway.

hannesm:
thx @gasche.  IMHO good to merge!

avsm:
ok, will cut a release shortly.

select_ppx_version.sh
```diff
@@ -1,10 +1,17 @@
 #!/bin/sh -e
 
+case $1 in
+  clean) rm -f ppx/ppx_cstruct.ml; exit 0 ;;
+  conf) (* continue *) ;;
```
gasche: looking at the patch again, this line uses OCaml comment syntax, which provokes the following behaviour from the root of the repository:

```$ (* continue *)
bash: appveyor.yml: command not found
```

(The first `*` expands to the list of files in alphabetic order and the first, `appveyor.ml`, gets called as a command)

It would be nice to commit a minor change to remove the comment or use proper shell comment syntax.

avsm: Yes, fixed this up in #112

PR #112, opened 2016-08-16
release 2.3.0, and add ocaml 4.04 to the test matrix

avsm:


gasche:
Could you send an `opam-repository` PR as well, so that user programs can be tested with the new release? Or is it already there and I missed it?

avsm:
It's not released in this repository yet until the tests pass -- Travis was blocked yesterday.

It looks like 4.04 is failing on REVDEPS, so this is ok to merge at this point.

PR #124, opened 2017-02-09
Correction to description of shift function

orbifx:


yomimono:
Thanks for this, @orbifx !

PR #125, opened 2017-03-03
Add OCaml 4.05 support

avsm:


avsm:
There is a suspicious lack of changes in the 4.04.0 to 4.05.0 PPX, so I just made this a symlink

avsm:
Passing base tests and then failing because no Lwt yet (https://github.com/ocsigen/lwt/issues/321).  This is ok to merge to unblock 4.05 support.

PR #126, opened 2017-03-06
Symlink 4.03 ppx to 4.04

rgrinberg:
The ppx for 4.03 is the same modulo typo

Here's the diff between the two:

```
<   (* Generate signaturs of the form {get/set}_<struct>_<field> *)
---
>   (* Generate signatures of the form {get/set}_<struct>_<field> *)
```


hannesm:
I'm not in favour of this, reasoning provided by @gasche in #111

rgrinberg:
But then why is 4.05 symlinked to 4.04 then? @avsm ?

avsm:
What's your preferred solution @hannesm? To modify the configure script?

I would like to move to use the ocaml-migrate-parsetree package in the medium term rather than have this version selection anyway...

hannesm:
@rgrinberg oh, wasn't aware of this.

@avsm the simplest for understanding is a file copy imho.  I agree that migrate-parsetree would be a good solution.

rgrinberg:
I understand gasche's argument as avoiding complicating the configure script is a good idea. But symlinking doesn't seem to be against that, it's just as "copy friendly". That being said, symlinks aren't windows friendly, so perhaps it should be avoided anyway.

avsm:
closing in favour of #127 which gets rid of allll of the things

PR #127, opened 2017-03-07
Port to ocaml-migrate-parsetree

let-def:
This is port of the project to [ocaml-migrate-parsetree](https://github.com/let-def/ocaml-migrate-parsetree). The library aims at making ppx portable accross compiler versions.

The API is likely to change in the upcoming days, so this is not ready for merging.


avsm:
Has this interface settled down now @let-def? We're interested in unblocking mirage compilation on OCaml 4.05 to try AFL -- cc @yomimono @talex5

let-def:
I think so. The library might receive a few update before 4.05, but the API will remain compatible.

let-def:
Ok, I think the final API made it to ocaml-migrate-parsetree. It has not been merged yet, but it would be nice to rerun tests after (I don't think I can?) and maybe consider merge.

PR: https://github.com/ocaml/opam-repository/pull/8772

avsm:
restarted Travis

let-def:
I am not sure how to express the opam constraints:

The package optionally depends on `ocaml-migrate-parsetree` and `ppx_tools_versioned`. If both are available, then the ppx can be built.

Since `ocaml-migrate-parsetree` is a direct dependency of `ppx_tools_versioned`, the former can probably be omitted, but that would be slighly underspecifying the constraints. Can you check that the opam file correctly specify the optional deps and the resulting conditional compilation?

avsm:
> Since ocaml-migrate-parsetree is a direct dependency of ppx_tools_versioned

It's fine to just depend on ppx_tools_versioned then, as long as we remember that we need to fix things up if the direct dependencies change in the future. That'll be picked up revdeps testing most likely even if we do forget.

avsm:
fails on 4.05 but thats expected due to Lwt for now -- thanks for this contribution!

PR #130, opened 2017-03-28
always enable ppx

let-def:


hannesm:
but what is the motivation for this change?  I don't understand, is cstruct without ppx now out of fashion / not used anywhere?

rgrinberg:
@hannesm without this change some downstream dependencies might cease to build becuse not everybody has the new depopt (namely ppx_tools.versioned and ocaml-migrate-parsetree) installed.

avsm:
@hannesm it's only a transitional change. the next round of changes will split up the cstruct repo into multiple OPAM packages with a separate ppx_cstruct, but this of course requires numerous revdeps changes.  So staging this one first.

PR #132, opened 2017-03-29
README: Use SVG build badge

olleolleolle:
This very simple change makes the build badge easier to read. Because SVG is shinier.


avsm:
Shiny! Thank you!

PR #133, opened 2017-03-31
port to jbuilder

avsm:
still need to do ppx and hook in tests, and ensure both standalone ppx work.

This PR does the work of splitting up PPX into several packages which are all standalone and independently installable through OPAM, so the core `cstruct` will not depend on ppx or lwt or async.

cc @rgrinberg


rgrinberg:
@avsm Mind letting me push on this branch? I'd like to proceed with this.

avsm:
Go for it!

rgrinberg:
@avsm OK I got some progress on this. I think you forgot to let me push on this branch though.

rgrinberg:
Things left to do
- [ ] what to do with cstruct.js? does it needs its own cstruct package? E.g. cstruct-js
- [ ] test out ppx compat. omp + classical ppx

cc @samoht @let-def

rgrinberg:
@let-def tested this work with ocamlbuild and jbuilder. no need to test but you can still have a glance if you'd like

rgrinberg:
@samoht @avsm for cstruct.js I just elected to keep everything the same and make it a part of cstruct directly. If there's a for these js stubs to have their dependencies, we can split it into its own package then.

I think this port to jbuilder is ready, so we can remove the WIP from the title and consider this for merging.

avsm:
Awesome, thanks @rgrinberg! I'll use this PR to setup a revdeps test on the cstruct repo to help with the upper bounds.

samoht:
Can you please rebase? I'll cut a release without jbuilder and then we can make a new release with jbuilder in.

rgrinberg:
Rebased @samoht

avsm:
Just debugging travis...

PR #136, opened 2017-04-25
Add missing dependency on ppx_tools_versioned

let-def:
This shoud fix the build with latest ppx_tools_versioned.


let-def:
(Without affecting the build with previous versions)

let-def:
The first fix was not fixing the root cause...
It is simply a dependency on `ppx_tools_versioned` that is missing. It used to be provided by `ppx_tools_versioned.metaquot`, but that was an unnecessary dependency that was removed in latest version.

rgrinberg:
@avsm or @samoht would it be possible to merge and release this already? I see this annoying quite a few users.

samoht:
Ok I'll do that today.

samoht:
Merging now: I'll fix the CI issues in an upcoming PR.

PR #137, opened 2017-05-03
Prepare the release

samoht:


samoht:
Remaining errors are transient errors (seems that TravisCI got a network outage..)

PR #138, opened 2017-05-11
release 3.0.0

avsm:
- add topkg-bridge also


pkg/pkg.ml
```diff
@@ -0,0 +1,8 @@
+#use "topfind"
+#require "topkg-jbuilder"
+
```
samoht: probably better to use `topkg-jbuilder.auto` and remove the rest of the file

avsm: it doesnt detect cstruct as the package name, possibly because ppx_cstruct breaks the rule of everything starting with `cstruct-` -- hence I set it manually. cc @dim

Issue #140, opened 2017-05-11
Crash in cstub for check_alignment_bigstring


cfcs:
When you pass a `0` argument to `check_alignment` the process crashes due to a division-by-zero error in the c stub.

.mli:
```ocaml
val check_alignment : t -> int -> bool
(** [check_alignment cstr alignment] is [true] if the first byte stored
    within [cstr] is at a memory address where [address mod alignment = 0],
    [false] otherwise.
    Typical uses are to check a buffer is aligned to a page or disk sector
    boundary. *)
```

cstruct.ml:
```ocaml
external check_alignment_bigstring : buffer -> int -> int -> bool = "caml_check_alignment_bigstring"

let check_alignment t alignment = check_alignment_bigstring t.buffer t.off alignment
```

cstruct_stubs.c:
```c
CAMLprim value
caml_check_alignment_bigstring(value val_buf, value val_ofs, value val_alignment)
{
  uint64_t address = (uint64_t) (Caml_ba_data_val(val_buf) + Long_val(val_ofs));
  int alignment = Int_val(val_alignment);
  return Val_bool(address % alignment == 0);
}
```

When `val_alignment` is `0` this crashes the program.

cfcs:
By the way, I am curious as to why it adds the `Long_val(val_ofs)`. Are there scenarios where t.off is not 0?

avsm:
Reproduced and test case added in 86216e878a3952993e512572419e33c7c4859d03

avsm:
> Are there scenarios where t.off is not 0

For multi-page cstructs the t.off might be >0 -- and we can use this function to test if the specific cstruct view is page-aligned with respect to the underlying memory.

cfcs:
Aha, okay!

PR #141, opened 2017-05-11
Expose pointer to Cstruct data

cfcs:
I needed this to talk to external C code from MirageOS.
This functionality already exists in Ctypes, but unfortunately that didn't seem to compile for me.

PR #142, opened 2017-05-11
check_alignment: Fix crash when alignment == 0

cfcs:
Fixes https://github.com/mirage/ocaml-cstruct/issues/140

I believe this should be done in OCaml, and that https://github.com/mirage/ocaml-cstruct/pull/141 makes that possible (note that the code below handles negative ints differently):
```ocaml
let check_alignment buf alignment : bool =
  match Cstruct.get_pointer buf with
  | None -> false
  | Some ptr -> try ptr mod alignment = 0 with Division_by_zero -> false
```

### DIFF
--- a/lib/cstruct_stubs.c
+++ b/lib/cstruct_stubs.c
@@ -76,5 +76,9 @@ caml_check_alignment_bigstring(value val_buf, value val_ofs, value val_alignment
 {
   uint64_t address = (uint64_t) (Caml_ba_data_val(val_buf) + Long_val(val_ofs));
   int alignment = Int_val(val_alignment);
+
+  if (0 >= alignment) // Avoid division by zero or % on negative ints
+    return Val_bool(0);
+
   return Val_bool(address % alignment == 0);
 }


cfcs:
BTW I don't know if I'm doing something wrong, but I couldn't get opam to install the package.
I had to do the following to get it building:
```bash
ln -s /bin/true ./configure
echo $'install:\n\tjbuilder install\n' >> Makefile
echo {$'\njs-',$'\njs-un'}$'install:\n\ttrue\n' >> Makefile
```

talex5:
I'd prefer it to crash than give a fake answer when given invalid input. Why are you passing zero here?

avsm:
Likewise I think that raising an exception here would be better (despite making check_alignment a little more expensive)

avsm:
see #145 for my preferred fix - does that look ok?

cfcs:
@avsm looks good! :-)

@talex5 Division by zero is undefined behavior in C. In this case (on my x86_64) it makes the unikernel crash without an error message, but it could do all sorts of other horrible things in other cases. Anyway, when Result is not an option (no pun intended), and we can't have dependent types, I prefer the debugging experience to be limited to having to deal with the unpleasant realm of OCaml exceptions rather than having to set up gdb with Xen to figure out what the fuck is going on. If you're feeling unconvinced, see https://en.wikipedia.org/wiki/Division_by_zero#Historical_accidents

avsm:
I think/hope/am-pretty-convinced that @talex5 meant "crash with an OCaml exception" as opposed to "depend on C undefined behaviour to crash" :-)

talex5:
Yes, indeed. Fix LGTM.

PR #145, opened 2017-05-11
check_alignment: ensure that alignment is a positive integer

### LINKED ISSUE #140: Crash in cstub for check_alignment_bigstring

avsm:
fixes #140

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -106,7 +106,7 @@ external check_alignment_bigstring : buffer -> int -> int -> bool = "caml_check_
 let check_alignment t alignment =
   if alignment > 0 then
     check_alignment_bigstring t.buffer t.off alignment
-  else raise (Invalid_argument "check_alignment must be positive integer")
+  else invalid_arg "check_alignment must be positive integer"
 
 type byte = char
 
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -103,7 +103,10 @@ let check_bounds t len =
 
 external check_alignment_bigstring : buffer -> int -> int -> bool = "caml_check_alignment_bigstring"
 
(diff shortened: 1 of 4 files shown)


lib/cstruct.ml
```diff
@@ -103,7 +103,10 @@ let check_bounds t len =
 
 external check_alignment_bigstring : buffer -> int -> int -> bool = "caml_check_alignment_bigstring"
 
-let check_alignment t alignment = check_alignment_bigstring t.buffer t.off alignment
+let check_alignment t alignment =
+  if alignment > 0 then
+    check_alignment_bigstring t.buffer t.off alignment
+  else raise (Invalid_argument "check_alignment must be positive integer")
```
rgrinberg: no love for `invalid_arg`?

avsm: fixed in 80428ed

PR #149, opened 2017-06-08
Problem chaining cstruct.ppx with lwt.ppx in jbuilder

g2p:
The PR contains a small test case to reproduce the bug.

Use

    opam pin add lwt https://github.com/andrewray/lwt.git#jbuild

to install a jbuilder-compatible lwt.

This works with cstruct 2.4.1.

### DIFF
--- a/reproducer/Makefile
+++ b/reproducer/Makefile
@@ -1,0 +1,2 @@
+build:
+	jbuilder build example.cmx
--- a/reproducer/example.ml
+++ b/reproducer/example.ml
@@ -1,0 +1,6 @@
+[%%cstruct type foo = {
+  magic: uint8_t [@len 16];
+}[@@little_endian]]
+
+let%lwt foo = Lwt.return ()
+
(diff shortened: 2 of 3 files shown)


avsm:
Cc @djs55 who is having a similar issue with sexp and cstruct

djs55:
I had 2 problems:

- `ppx_cstruct` invoking the `main` function itself, rather than letting `jbuilder` call it after allowing all the drivers to register-- see #150 
- an overly-aggressive attribute check expecting there to be no unprocessed attributes after every pass (?) -- I've worked around this with `-no-check`: https://github.com/mirage/ocaml-vchan/blob/1b39c3ae46f96c02deb2ad33e480b7301bf5fe4e/lib/jbuild#L6

g2p:
With #150 applied, the alias cstruct.ppx doesn't work, and doesn't accept a -no-check option, but ppx_cstruct works okay.

Issue #148, opened 2017-06-08
ppx not working with lwt


andrewray:
Is it correct to have the file standalone.ml built into the ppx library for jbuilder (as it appears to be)?

The lwt syntax extension isn't playing nice with cstruct.  

cc @diml

diml:
Nope, it shouldn't be there indeed. I was actually looking at this as well as someone asked me about it. There is a fix in #150

andrewray:
Ta.  Closing as addressed in #149 and #150

PR #151, opened 2017-06-08
Attempt to fix composition of lwt and cstruct

avsm:
combines test cases with #148 #149 #150 but its not clear that it's all fixed yet


avsm:
@djs55 does `make test` in this PR succeed for you? fails for me

djs55:
I get
```
$ make test
jbuilder runtest
Error: Conflict between internal and external version of library "ppx_cstruct":
- it is defined locally in ppx/jbuild
- it is required by external library "cstruct.ppx"
- external library "cstruct.ppx" is required in ppx_test/with-lwt/jbuild (context default)
This cannot work.
make: *** [test] Error 1
```

djs55:
Even if I comment out all the `cstruct.ppx` from that example I still get a failure:
```
(cd _build/default && /usr/local/bin/ocamlc.opt -w -40 -g -bin-annot -I /Users/djs/.opam/system/lib/bytes -I /Users/djs/.opam/system/lib/lwt -I /Users/djs/.opam/system/lib/ocplib-endian -I /Users/djs/.opam/system/lib/result -I /Users/djs/.opam/system/lib/sexplib -I /Users/djs/.opam/system/lib/sexplib/0 -I /usr/local/lib/ocaml -I lib -no-alias-deps -I ppx_test/with-lwt -o ppx_test/with-lwt/ppx_cstruct_and_lwt.cmo -c -impl ppx_test/with-lwt/ppx_cstruct_and_lwt.pp.ml)
File "ppx_test/with-lwt/ppx_cstruct_and_lwt.ml", line 6, characters 4-7:
Error: Uninterpreted extension 'lwt'.
make: *** [test] Error 1
```

djs55:
I've pushed patches which make it work for me but only if I pin `master` of `lwt`. With `lwt.3.0.0` I get "Uninterpreted extension 'lwt'".

The main change I made is to use `ppx_cstruct` in the example rather than `cstruct.ppx`. I think `cstruct.ppx` should still work for other packages, but I think `jbuilder` is confused about it since it only exists in the `META.cstruct.template` file. Perhaps if we actually made it into a first-class (but empty) library it would work more naturally?

g2p:
I wouldn't expect lwt.3.0.0 to work, as the lwt.ppx extension hasn't been migrated to ocaml-migrate-parsetree.

djs55:
@g2p great, thanks for the clarification!

avsm:
i've updated changes, and added all the right testcases i think, so this can be released once it goes through ci

PR #155, opened 2017-06-20
Fix compilation of ppx_test

samoht:
/cc @avsm @djs55

### DIFF
--- a/ppx_test/with-sexp/jbuild
+++ b/ppx_test/with-sexp/jbuild
@@ -2,11 +2,10 @@
 
 (executables
  ((names (ppx_cstruct_and_sexp))
-  (preprocess (pps (cstruct.ppx ppx_sexp_conv -no-check)))
+  (preprocess (pps (cstruct_ppx ppx_sexp_conv -no-check)))
   (libraries (cstruct))))
 
 (alias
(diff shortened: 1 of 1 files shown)


djs55:
I'm a bit confused -- where does `cstruct-ppx` come from?

samoht:
Fixed to match https://github.com/mirage/ocaml-cstruct/blob/master/ppx/jbuild#L5

avsm:
that should be ppx_cstruct

ppx_test/with-sexp/jbuild
```diff
@@ -2,11 +2,10 @@
 
 (executables
  ((names (ppx_cstruct_and_sexp))
-  (preprocess (pps (cstruct.ppx ppx_sexp_conv -no-check)))
+  (preprocess (pps (cstruct_ppx ppx_sexp_conv -no-check)))
```
avsm: dont release with this -- it should be ppx_cstruct

samoht: woops. Why does this pass on Travis?

samoht: See https://github.com/mirage/ocaml-cstruct/pull/156

PR #158, opened 2017-07-07
Fixes: performance regression + opam files

hannesm:
...maybe we need more code reviews?   or slower merges?


hannesm:
appveyor seems to be unhappy... any jbuilder experts around?
```
[cstruct: jbuilder runtest] Command started
+ jbuilder "runtest" "-p" "cstruct" (CWD=C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/cstruct.3.0.2)
+ jbuilder "runtest" "-p" "cstruct" (CWD=C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/cstruct.3.0.2)
- Error: I couldn't find 'ocaml-migrate-parsetree.driver-main'.
- I need this library in order to use ppx rewriters.
- See the manual for details.
- Hint: opam install ocaml-migrate-parsetree
[ERROR] The compilation of cstruct failed at "jbuilder runtest -p cstruct".
```

avsm:
Where did this regress from?

hannesm:
see commit message, #102

hannesm:
travis is happy after fixing up yet another glitch in the opam file!

ppx_cstruct.opam
```diff
@@ -23,5 +23,6 @@ depends: [
   "ocaml-migrate-parsetree"
   "ppx_driver"    {test & >= "v0.9.0"}
   "ppx_sexp_conv" {test}
+  "cstruct-unix"  {test}
```
avsm: This wasn't caught due to a missing `TESTS=true` in the travis.yml -- any chance you could add that so it runs on the CI as well?

avsm:
To clarify the purpose of this change, it is to use the `create_unsafe` interface to _internally_ allocate buffers that are fully initialised by their respective methods.  I've looked at each method and satisfied myself that it is only used in functions that subsequently overwrite the contents of the buffer.

cc @samoht @talex5 @djs55 to look at the change (@yomimono is on vacation)

ppx_cstruct.opam
hannesm: where in travis.yml needs that `TEST=true` to be?  this didn't use to be the case, or was it?  I remember having seen travis executing tests without it!?!?

samoht: I think there are different default for the "classic" TravisCI scripts vs. the Docker-based ones. The classic ones have TEST=true by default.

avsm: Yeah, I'm not sure why the container-based ones do not run it by default. I explicitly add it when TESTS are desired (some repositories have non-terminating tests, as older tcpip revisions)

avsm: Also it's useful to have some of the matrix _not_ run tests, so you can verify that the package installs without test dependencies.

hannesm: @samoht @avsm thx for clarification, I added this below.  I'm surprised by the sudden change of defaults between different travis flavours.

hannesm:
On 07/07/2017 14:31, Anil Madhavapeddy wrote:
> Also it's useful to have some of the matrix _not_ run tests, so you can verify that the package installs without test dependencies.

that's always the case!  in tls I use `TESTS=false` in parts of the matrix.

cstruct.opam
```diff
@@ -12,7 +12,7 @@ build: [
   ["jbuilder" "subst" "-p" name "--name" name] {pinned}
   ["jbuilder" "build" "-p" name "-j" jobs]
 ]
-build-depends: ["jbuilder" "runtest" "-p" name]
+build-test: ["jbuilder" "runtest" "-p" name]
```
avsm: This is causing the appveyor failure, since the ppx tests are being built when only cstruct is installed. It should be fixable by assigning the `ppx_test` directory to the ppx_cstruct package with something like:

```




(executables
  (package ppx_cstruct)
  ((libraries (cstruct-unix oUnit))
  (preprocess (pps (ppx_cstruct)))
  (names (pcap basic enum))))
```

(note the new `package` field to manually assign it rather than defaulting to `cstruct`).

This should also fail on Linux, but our test coverage isn't quite there yet I think.

hannesm: I suspect you can directly push to this branch if you want to fix it.  I feel like I already fixed enough regressions today (within this PR, just look at the commit messages).

lib/cstruct.ml
```diff
@@ -410,16 +410,16 @@ let rec fold f next acc = match next () with
...
   let l1 = len cs1 and l2 = len cs2 in
-  let cs = create (l1 + l2) in
+  let cs = create_unsafe (l1 + l2) in
   blit cs1 0 cs 0  l1 ;
   blit cs2 0 cs l1 l2 ;
   cs
 
 let concat = function
-  | []   -> create 0
+  | []   -> create_unsafe 0
```
talex5: Is it worth creating a `Cstrust.empty` value for this? I've wanted one a few times in other projects.

cstruct.opam
avsm: Sure, I'll fix it once the actual code is sorted. 

To address some of your queries about more reviews and regressions: it's not actually a regression as we were never actually testing some of this functionality before.  Breaking out ppx into a separate package  is a new feature that is causing the need for separating the test dependencies as well -- a feature I implemented upon a request from you in fact :-)

The original PR in question had 3 reviews, so the decision then was to prioritise safety over performance.  The decision now is to restore performance but with careful review to ensure that there is no safety regression.

Iow, I see a healthy process in place with features being added, reviews being conducted, and increasing test coverage over a more modular codebase that supports separate dependencies across ppx, async, lwt, unix and mirage.  Thanks for taking the time to submit these ancillary fixes as part of this process!

lib/cstruct.ml
avsm: i've opened a separate issue for this in https://github.com/mirage/ocaml-cstruct/issues/163

PR #159, opened 2017-07-07
Fix an arithmetic overflow bug (lenv, copyv)

yallop:
Modular arithmetic strikes again.

Here's a demonstration on a 32-bit system.

Before:

```ocaml
# let b = Cstruct.create max_int and c = Cstruct.create 3 ;;
val b : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 1073741823}
val c : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 3}
# Cstruct.lenv [b; b; c];;
- : int = 1
# Cstruct.copyv [b; b; c];;
Segmentation fault
```

After:

```ocaml
# let b = Cstruct.create max_int and c = Cstruct.create 3 ;;
val b : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 1073741823}
val c : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 3}
# Cstruct.lenv [b; b; c];;
Exception: Invalid_argument "Cstruct.lenv".
# Cstruct.copyv [b; b; c];;
Exception: Invalid_argument "Cstruct.copyv".
```

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -296,13 +296,22 @@ end
 let len t =
   t.len
 
-let lenv = function
-  | []  -> 0
-  | [t] -> len t
-  | ts  -> List.fold_left (fun a b -> len b + a) 0 ts
+(** [sum_lengths ~caller acc l] is [acc] plus the sum of the lengths
+    of the elements of [l].  Raises [Invalid_argument caller] if 
+    arithmetic overflows. *)
+let rec sum_lengths_aux ~caller acc = function
+  | [] -> acc
+  | h :: t -> 
+     let sum = len h + acc in
+     if sum < acc then invalid_arg caller
+     else sum_lengths_aux ~caller sum t
+
+let sum_lengths ~caller l = sum_lengths_aux ~caller 0 l
+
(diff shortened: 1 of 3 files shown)


lib/cstruct.ml
```diff
@@ -296,13 +296,23 @@ end
...
-let lenv = function
-  | []  -> 0
-  | [t] -> len t
-  | ts  -> List.fold_left (fun a b -> len b + a) 0 ts
+(** [sum_lengths ~caller acc l] is [acc] plus the sum of the lengths
+    of the elements of [l].  Raises [Invalid_argument caller] if 
+    arithmetic overflows. *)
+let rec sum_lengths_aux ~caller acc = function
+  | [] -> acc
+  | h :: [] -> len h + acc
```
talex5: How do we know this case doesn't overflow?

yallop: It can overflow, but the result will be negative if it does, and subsequently caught by the caller.  But this is probably too subtle and we should just check directly here.

yallop: In fact, we should certainly check here, since `lenv` won't catch the overflow.

PR #160, opened 2017-07-10
Reject negative destination offsets in 'blit'

yallop:
Before:

```
# Cstruct.blit (Cstruct.create 0xf) 0 (Cstruct.create 0xf) (-0xf) 0xf;;
*** Error in `/home/jeremy/.opam/4.04.2/bin/ocamlrun': free(): invalid next size (fast): 0x0000561b7e0b0470 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x70bcb)[0x7f10fe2f9bcb]
/lib/x86_64-linux-gnu/libc.so.6(+0x76f96)[0x7f10fe2fff96]
[...]
```

After:

```
# Cstruct.blit (Cstruct.create 0xf) 0 (Cstruct.create 0xf) (-0xf) 0xf;;
Exception:
Invalid_argument
 "Cstruct.blit src=[0,15](15) dst=[0,15](15) dst-off=-15 len=15".
```

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -169,7 +169,7 @@ let copy src srcoff len =
 let blit src srcoff dst dstoff len =
   if len < 0 || srcoff < 0 || src.len - srcoff < len then
     err_blit_src src dst srcoff len
-  else if dst.len - dstoff < len then
+  else if dstoff < 0 || dst.len - dstoff < len then
     err_blit_dst src dst dstoff len
   else
     unsafe_blit_bigstring_to_bigstring src.buffer (src.off+srcoff) dst.buffer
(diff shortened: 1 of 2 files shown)


hannesm:
this looks good to me -- there's a small inconsistency though: `blit_from_string`, `blit_from_bytes`, and `blit_to_bytes` do a `dstoff < 0` check, but in the first conditional, calling to the `err_blit_X_src`.

I agree that the solution in here is correct, and maybe we should adapt it to the other `blit_*` functions as well (it's mainly cosmetic AFAICT)!?

talex5:
LGTM, but can we get some AFL fuzz tests for this library? I suspect it would find all of these bugs in seconds.

samoht:
> (I can add some crowbar tests if you want)

That's a great idea!

talex5:
(fuzzing added in #164)

PR #164, opened 2017-07-11
Add some fuzz tests

talex5:
When running the tests individually, it:

- Found https://github.com/mirage/ocaml-cstruct/pull/160 in 17s.
- Found an overflow in of_bigarray in 1s (fixed).
- Found an overflow in sub in 1s (fixed).

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -85,7 +85,7 @@ let of_bigarray ?(off=0) ?len buffer =
     match len with
     | None     -> dim - off
     | Some len -> len in
-  if off < 0 || len < 0 || off + len > dim then err_of_bigarray off len
+  if off < 0 || len < 0 || off + len < 0 || off + len > dim then err_of_bigarray off len
   else { buffer; off; len }
 
 let to_bigarray buffer =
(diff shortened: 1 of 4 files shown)


avsm:
Awesome!!! Thanks for this.

lib/cstruct.ml
```diff
@@ -96,7 +96,7 @@ let create_unsafe len =
   { buffer ; len ; off = 0 }
 
 let check_bounds t len =
-  Bigarray.Array1.dim t.buffer >= len
+  len >= 0 && Bigarray.Array1.dim t.buffer >= len
```
yallop: @talex5: I think the docstring for `check_bounds` needs to be updated to mention the check that `len` is positive.

talex5: Maybe the public function should raise `Invalid_argument` in this case. It's all a bit of a mess. Possibly we should just redefine `+` to check for overflow in modules which use unsafe accessors.

yallop: Yes, `Invalid_argument` would be reasonable.  I'm not sure how the public function is intended to be used: it uses the size of the underlying buffer, not the length of the cstruct itself, which is a bit different from most functions in the API.

It'd be quite handy to have arithmetic operations that raise on overflow in the standard library.

hannesm: > It'd be quite handy to have arithmetic operations that raise on overflow in the standard library.

the base for this was already done in https://github.com/ocaml/ocaml/pull/1060 (merged in trunk, not part of 4.05) -- but these C symbols (esp. add/sub) are not yet used anywhere.

avsm: I don't think check_bounds needs to be public any more. It was originally because it was inserted by a camlp4 extension, but nowadays it's just done inline in the library itself.

Issue #154, opened 2017-06-16
"noalloc" deprecation warnings when building


cfcs:
FWIW

```
File "lib/cstruct.ml", line 160, characters 0-138:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 162, characters 0-132:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 164, characters 0-132:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 166, characters 0-132:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 168, characters 0-132:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 170, characters 0-117:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
File "lib/cstruct.ml", line 172, characters 0-102:
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
```

avsm:
Thanks -- we need to bump up the minimum supported version of OCaml to 4.03.0, which is reasonable nowadays given that the rest of MirageOS' support requires that as well.

avsm:
fixed in #157 and will be in 3.1.0

PR #165, opened 2017-07-11
start preparing 3.1.0

avsm:
includes #164 #160 #159

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ v3.1.0 2017-07-11
 - Add AFL fuzz tests using Crowbar, which independently discovered
   #160 and also an overflow in `of_bigarray` and `sub`, now bith
   fixed (#164 by @talex5)
+- Improve performance of several allocation functions by eliminating an
+  unnecessary buffer zero step (#158)
 - Bump the required minimum OCaml version up to 4.03.0.
 
 v3.0.2 2017-06-14
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,13 @@
+v3.1.0 2017-07-11
+-----------------
+
+- Fix arithmetic overflow in `Cstruct.lenv` and `copyv` (#159 by @yallop)
+- Reject negative destination offsets in `blit` (#160 by @yallop)
+- Add AFL fuzz tests using Crowbar, which independently discovered
+  #160 and also an overflow in `of_bigarray` and `sub`, now bith
+  fixed (#164 by @talex5)
+- Bump the required minimum OCaml version up to 4.03.0.
+
 v3.0.2 2017-06-14
 -----------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,7 +7,7 @@ v3.1.0 2017-07-11
   #160 and also an overflow in `of_bigarray` and `sub`, now bith
   fixed (#164 by @talex5)
 - Improve performance of several allocation functions by eliminating an
-  unnecessary buffer zero step (#158)
+  unnecessary buffer zero step (#158 by @hannesm)
 - Bump the required minimum OCaml version up to 4.03.0.
 
 v3.0.2 2017-06-14
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all clean
+.PHONY: all clean fuzz build-fuzz
 
 all:
 	jbuilder build
@@ -8,6 +8,14 @@ clean:
 
 test:
 	jbuilder runtest --dev
+
+build-fuzz:
+	jbuilder build --dev fuzz/fuzz.exe
+
+fuzz: build-fuzz
+	mkdir -p _build/in
+	echo > _build/in/empty
+	afl-fuzz -i _build/in -o _build/out -- _build/default/fuzz/fuzz.exe @@
 
 REPO=../../mirage/opam-repository
 PACKAGES=$(REPO)/packages
--- a/cstruct-async.opam
+++ b/cstruct-async.opam
@@ -2,7 +2,7 @@ opam-version: "1.2"
 maintainer:   "anil@recoil.org"
 authors:      ["Anil Madhavapeddy" "Richard Mortier" "Thomas Gazagnaire"
                "Pierre Chambart" "David Kaloper" "Jeremy Yallop" "David Scott"
-               "Mindy Preston" ]
+               "Mindy Preston" "Thomas Leonard" ]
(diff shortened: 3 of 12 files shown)


hannesm:
- you're missing to mention #158 in changes
- what is the reason to require 4.03 here?  cstruct is a library used in a lot of projects, bumping the required OCaml version should be done carefully (since you require all reverse dependencies to either stay away from updates (which with the current changes looks like a bad idea, since several potential segmentation faults are fixed), or to require 4.03 as well - which has not yet been adapted by all stable distribution channels (i.e. ubuntu 16.10 still ships 4.02.3 AFAICT)

avsm:
> what is the reason to require 4.03 here? 

I'm bumping all the MirageOS libraries to 4.03 to be consistent, since this was the consensus for the MirageOS 3 release. Our test matrix is currently unreasonably large, and we are running into a number of deprecated warnings. Furthermore, dependencies like AFL/Crowbar only work on modern versions, and the switch to attributes for things like `noalloc` ( #154) is risky since it'll be silently accepted in older versions.

So to turn this around -- is there a concrete reason you need 4.02 and its associated support burden?

hannesm:
On 11/07/2017 13:44, Anil Madhavapeddy wrote:
> So to turn this around -- is there a concrete reason you need 4.02
> and its associated support burden?

as mentioned, there are people who want to "try out OCaml" or use a
single application (e.g.
[jackline](https://github.com/hannesm/jackline)), who use a stable
distribution (and yes, there are jackline users who use 4.02 and (try
to) use opam 1.2.0 (which unfortunately transitioned into a
not-too-usable-state with the cstruct 3.0.0 release).

I'm all for deprecating 4.02 (and maybe even 4.03 now that 4.05 is
nearly out) for MirageOS, but with cstruct (same for ipaddr, uri) I'd be
more conservative, as these are libraries used as well outside of the
MirageOS ecosystem.

avsm:
>  (try to) use opam 1.2.0 (which unfortunately transitioned into a not-too-usable-state with the cstruct 3.0.0 release).

I'm afraid we just can't support older software indefinitely. opam 1.2.0 is most definitely deprecated now -- http://opam.ocaml.org/blog/deprecating-opam-1-2-0/

> I'm all for deprecating 4.02 (and maybe even 4.03 now that 4.05 is nearly out) for MirageOS, but with cstruct (same for ipaddr, uri) I'd be more conservative, as these are libraries used as well outside of the MirageOS ecosystem.

And similarly, cohttp and many of the other libraries are also used elsewhere.  We just don't have the resources in the current level of automation and testing to effectively support 4 parallel releases (4.02/3/4/5 round the corner).  The current choice is better improving quality in our latest releases (e.g. adding and using afl pervasively) _vs_ maintaining backwards compatibility, and we have to chose the former in order to keep moving forward.
 
I think we can get to the point where an LTS stream is possible, but for now it would be better to focus on getting Jackline users a set of installation instructions that let them use the latest tools (opam 1.2.2 and ocaml 4.03.0 especially). If you let me know which distro is lagging, that'd be appreciated...

hannesm:
On 11/07/2017 13:59, Anil Madhavapeddy wrote:
> And similarly, cohttp and many of the other libraries are also used
> elsewhere.  We just don't have the resources in the current level of
> automation and testing to effectively support 4 parallel releases
> (4.02/3/4/5 round the corner).  The current choice is better
> improving quality in our latest releases (e.g. adding and using afl
> pervasively) _vs_ maintaining backwards compatibility, and we have to
> chose the former in order to keep moving forward.

I don't think this library needs any effort to keep 4.02 compat, not
modifying the "available" in opam is all I ask for ;)

> I think we can get to the point where an LTS stream is possible, but
> for now it would be better to focus on getting Jackline users a set
> of installation instructions that let them use the latest tools (opam
> 1.2.2 and ocaml 4.03.0 especially). If you let me know which distro
> is lagging, that'd be appreciated...

Ubuntu 16.10 AFAICS, Whonix 13, FreeBSD - all of them still have 4.02.3
in their package system.

avsm:
> I don't think this library needs any effort to keep 4.02 compat, not modifying the "available" in opam is all I ask for ;)

For a start, we haven't been testing it in Travis, nor will we be fuzz testing it, nor can we move the codebase forward (#154).

> Ubuntu 16.10 AFAICS, Whonix 13, FreeBSD

All of these have opam 1.2.2 though (phew!), so why not bump Jackline's minimum version to 4.03.0 and request users to compile a static binary via a switch?  Could put a recipe in the repository to construct a `musl+4.03` switch for this purpose.

My concern here is for the users -- if you try to actually use 4.02.3 then there are a surprising number of downgrades pinning users back. Similar story for 4.01.0 as we did a staggered set of dependencies on 4.02.3.  We can fix this for future releases, but right now the only recommended way of installing opam libraries is to use opam 1.2.2+, and that supports --switch specifically to let us control the compiler platform more easily.  If you see people packaging up binary distributions of `cstruct` then we will need to take action to support a backport stream...

hannesm:
On 11/07/2017 14:15, Anil Madhavapeddy wrote:
> For a start, we haven't been testing it in Travis,

until recently (just a few days/hours ago) we did, or?

> nor can we move the codebase forward (#154).

hmm, I guess we misunderstand each other:  if we move the codebase to
use 4.03+ only features, we should increase the ocaml version.  but as
long as we don't use any 4.03+ features, why should it be increased?

>> Ubuntu 16.10 AFAICS, Whonix 13, FreeBSD
> 
> All of these have opam 1.2.2 though (phew!)

Whonix 13 still has opam 1.2.0 AFAICT.

avsm:
- not testing it since 48d260362436331b265b90ad3dcac772d6401776 -- not sure what version that is.  Certainly not been using it with 4.02 on MirageOS builds for quite some time.

- I want to fix #154 in this release and turn on strict warnings.

- Whonix13 needs fixing then if its using opam 1.2.0. This is absolutely not supported and really will be a terrible user experience. Would you have any contacts there who might be able to do a backport?

hannesm:
On 11/07/2017 14:25, Anil Madhavapeddy wrote:
> - Whonix13 needs fixing then if its using opam 1.2.0. This is absolutely not supported and really will be a terrible user experience. Would you have any contacts there who might be able to do a backport?

I don't.  Sorry.

avsm:
No problem -- I've posted to the forum to see if anyone else does: https://discuss.ocaml.org/t/rfc-deprecating-opam-1-2-0/332/10

avsm:
I'm going to ahead and merge this, and we can consider 4.02.3 backports as a separate issue if we can't migrate any existing users.

PR #177, opened 2017-10-01
retire ocplib-endian.  Since OCaml 4.01 the primitives are part of Bigarray

hannesm:
in the same spirit as https://github.com/inhabitedtype/faraday/pull/24 (thanks @seliopou) we can directly use the primitives provided by the OCaml runtime (since 4.01, https://caml.inria.fr/mantis/view.php?id=5771 -- cstruct is only available on `>= 4.02.3`).

Why?  Well, less dependencies are always better.  ocplib-endian is additionally LGPL licensed (see https://github.com/OCamlPro/ocplib-endian/blob/master/COPYING.txt) - thus pretty deep down in our dependency chain a not very permissively licensed library.

~~What do we loose? big endian support (which AFAICT hasn't been used anyways, or has it?) -- it can be re-added by either some preprocessing (cpp0?) or by dynamic checks for `Sys.big_endian`.  I don't know enough about cpp0 do go that way, but I can add the dynamic `Sys.big_endian` checks if this is desired.~~

EDIT: I added `Sys.big_endian` tests, performance could be improved using a preprocessor (but since ocplib as well uses runtime checks, there's no performance regression in here)

### DIFF
--- a/cstruct.opam
+++ b/cstruct.opam
@@ -16,7 +16,6 @@ build-test: ["jbuilder" "runtest" "-p" name]
 
 depends: [
   "jbuilder" {build & >="1.0+beta10"}
-  "ocplib-endian"
   "sexplib"
   "base-bytes"
   "ounit" {test}
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -223,74 +223,82 @@ let create len =
 
 let set_uint8 t i c =
   if i >= t.len || i < 0 then err_invalid_bounds "set_uint8" t i 1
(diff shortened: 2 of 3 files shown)


hannesm:
locally I observed breakage in `git`, fixed in https://github.com/mirage/ocaml-git/pull/233

avsm:
lgtm; i've started the git constraining PR to prepare this for release. I believe that Sys.big_endian can be inlined if it's sufficiently at the toplevel that the inliner picks it up. It might be worth experimenting with the conditional into a toplevel value and looking at the lambda output to see if it's still there after compilation.

hannesm:
is there anything I can do to move this PR forward?

PR #179, opened 2017-10-02
provide of_hex : string -> t

### LINKED ISSUE #178: implement of_hex (and to_hex?)
this would be a very convenient function to have, discussed further https://github.com/mirage/ocaml-cstruct/pull/175#issuecomment-328336725

since there are several places (such as [nocrypto](https://github.com/mirleft/ocaml-nocrypto/blob/v0.5.4/src/uncommon.ml#L211), [git](https://github.com/mirage/ocaml-git/blob/1.11.2/src/git/hash.ml#L153), [salsa20-core](https://github.com/abeaumont/ocaml-salsa20-core/blob/0.3.0/utils.ml#L1), [dns](https://github.com/mirage/ocaml-dns/blob/v1.0.0/lib/loader.ml#L282), [hex](https://github.com/mirage/ocaml-hex/blob/v1.1.1/lib/hex.ml#L59)) which implement exactly this function and depend on cstruct.

hannesm:
this is simply stolen from nocrypto, all credits go to @pqwy

closes #178 -- `to_hex` is already implemented (in the means of `hexdump_to_buffer` or `hexdump_pp`)

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -370,6 +370,39 @@ let of_bytes ?allocator buf =
     blit_from_bytes buf 0 c 0 buflen;
     set_len c buflen
 
+let of_hex str =
+  let string_fold ~f ~z str =
+    let st = ref z in
+    ( String.iter (fun c -> st := f !st c) str  ; !st )
+  in
+  let hexdigit p = function
+    | 'a' .. 'f' as x -> int_of_char x - 87
+    | 'A' .. 'F' as x -> int_of_char x - 55
+    | '0' .. '9' as x -> int_of_char x - 48
+    | x ->
+      Format.ksprintf invalid_arg "of_hex: invalid character at pos %d: %C" p x
+  in
+  let whitespace = function
+    | ' ' | '\t' | '\r' | '\n' -> true
+    | _ -> false
+  in
+  match
+    string_fold
+      ~f:(fun (cs, i, p, acc) ->
+          let p' = succ p in
+          function
+          | char when whitespace char -> (cs, i, p', acc)
+          | char ->
+            match acc, hexdigit p char with
+            | (None  , x) -> (cs, i, p', Some (x lsl 4))
+            | (Some y, x) -> set_uint8 cs i (x lor y) ; (cs, succ i, p', None))
+      ~z:(create_unsafe (String.length str lsr 1), 0, 0, None)
+      str
+  with
+  | _ , _, _, Some _ ->
+    Format.ksprintf invalid_arg "of_hex: odd numbers of characters"
+  | cs, i, _, _ -> sub cs 0 i
+
 let hexdump_pp fmt t =
   Format.pp_open_box fmt 0 ;
   for i = 0 to len t - 1 do
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -221,6 +221,15 @@ val of_bytes: ?allocator:(int -> t) -> Bytes.t -> t
     with the underlying buffer allocated by [alloc]. If [allocator] is not
     provided, [create] is used. *)
 
+val of_hex: string -> t
+(** [of_hex str] is the cstruct [cs].  Every pair of hex-encoded characters in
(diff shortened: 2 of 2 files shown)


lib/cstruct.ml
```diff
@@ -370,6 +370,35 @@ let of_bytes ?allocator buf =
...
+let string_fold ~f ~z str =
+  let st = ref z in
+  ( String.iter (fun c -> st := f !st c) str  ; !st )
+
+let of_hex str =
+  let hexdigit = function
+    | 'a' .. 'f' as x -> int_of_char x - 87
+    | 'A' .. 'F' as x -> int_of_char x - 55
+    | '0' .. '9' as x -> int_of_char x - 48
+    | x               -> Format.ksprintf invalid_arg "of_hex: `%c'" x
```
cfcs: May I suggest using `%C` here instead (escaping the argument),
alternatively `'%02x'" (int_of_char x)` ?
The latter will make it easier to spot the byte when trying to correlate with a `Cstruct.hexdump_pp`.

lib/cstruct.ml
```diff
@@ -370,6 +370,35 @@ let of_bytes ?allocator buf =
...
+  let whitespace = function
+    | ' ' | '\t' | '\r' | '\n' -> true
+    | _                        -> false
+  in
+  match
+    string_fold
+      ~f:(fun (cs, i, acc) -> function
+          | char when whitespace char -> (cs, i, acc)
+          | char ->
+            match (acc, hexdigit char) with
```
cfcs: I think printing the offset where the invalid byte was spotted (if it is invalid) would also be a good addition for debugging. For example by passing `i` to `hexdigit` here.

lib/cstruct.ml
```diff
@@ -370,6 +370,35 @@ let of_bytes ?allocator buf =
...
+            | (None  , x) -> (cs, i, Some (x lsl 4))
+            | (Some y, x) -> set_uint8 cs i (x lor y) ; (cs, succ i, None))
+      ~z:(create_unsafe (String.length str lsr 1), 0, None)
+      str
+  with
+  | (_ , _, Some _) -> Format.ksprintf invalid_arg "of_hex: dangling nibble"
```
cfcs: I had to look up `nibble`. It may be that I'm just not very educated, but I feel like `of_hex: uneven number of hex characters` would be easier to understand.

lib/cstruct.mli
```diff
@@ -221,6 +221,15 @@ val of_bytes: ?allocator:(int -> t) -> Bytes.t -> t
     with the underlying buffer allocated by [alloc]. If [allocator] is not
     provided, [create] is used. *)
 
+val of_hex: string -> t
+(** [of_hex str] is the cstruct of [str].  Every pair of hex-encoded characters
+    from [str] are converted to a single byte in the cstruct.  The resulting
+    cstruct is exactly half the size of [str].  Whitespaces (space, newline,
```
cfcs: When the [str] len is odd, that is not the case :) Maybe *The length of the resulting cstruct is half of the length of the hex characters in [str]*? (While technically the underlying buffer is longer (and contains unsanitized heap memory), you do return a substring for which it *is* true)

lib/cstruct.ml
hannesm: done - keep in mind `i` is the index into the cstruct, I introduced `p` for position in the string, and output this on error

lib/cstruct.ml
hannesm: see above, it is now printed

lib/cstruct.ml
hannesm: removed any mention of `nibble`

lib/cstruct.mli
hannesm: reworded

hannesm:
is there anything I can do to move this PR forward?

djs55:
A very useful function! Sorry for the delay. I need a better system for responding to PRs :(

PR #183, opened 2017-11-15
add "to_bytes : t -> bytes".  the reasoning behind this is that

hannesm:
(a) safe-string by default in 4.06 which demand us to separate string and bytes
(b) to_string is implemented as Bytes.to_string |> to_bytes
(c) read and write from stdlib now receive bytes
(d) user code otherwise has to: Cstruct.to_string |> Bytes.of_string (which copies)

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -339,12 +339,15 @@ let fillv ~src ~dst =
   aux dst 0 src
 
 
-let to_string t =
+let to_bytes t =
   let sz = len t in
   let b = Bytes.create sz in
   unsafe_blit_bigstring_to_bytes t.buffer t.off b 0 sz;
+  b
+
+let to_string t =
   (* The following call is safe, since b is not visible elsewhere. *)
-  Bytes.unsafe_to_string b
+  Bytes.unsafe_to_string (to_bytes t)
 
 let of_string ?allocator buf =
   let buflen = String.length buf in
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -346,7 +346,8 @@ let to_bytes t =
   b
 
 let to_string t =
-  (* The following call is safe, since b is not visible elsewhere. *)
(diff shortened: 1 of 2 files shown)


hannesm:
IMHO we should get this out and released before people start porting code to 4.06 which uses the `to_string |> Bytes.of_string` all over. (#177 and #179 should be merged before release as well, but this is just my opinion)

hannesm:
a question I have: what is the difference between `Bytes.t` and `bytes`? since when is the latter available in OCaml?

lib/cstruct.ml
```diff
@@ -339,12 +339,15 @@ let fillv ~src ~dst =
...
 
-let to_string t =
+let to_bytes t =
   let sz = len t in
   let b = Bytes.create sz in
   unsafe_blit_bigstring_to_bytes t.buffer t.off b 0 sz;
+  b
+
+let to_string t =
   (* The following call is safe, since b is not visible elsewhere. *)
```
yallop: `since b is not visible elsewhere` ↝ `since this is the only reference to the freshly-created value built by [to_bytes t]`.

hannesm:
https://github.com/mirage/mirage-flow/pull/38 is an example where this functionality would be great to have

djs55:
I think the `bytes` ocamlfind package containing `module Bytes` is a backwards compatibility shim:
```
s$ cat $(ocamlfind query bytes)/META
name="bytes"
version="[distributed with OCaml 4.02 or above]"
description="dummy backward-compatibility package for mutable strings"
requires=""
```

I also found https://caml.inria.fr/distrib/ocaml-4.02/notes/Changes
> OCaml 4.02.0 (29 Aug 2014):
> ---------------------------
>
> (Changes that can break existing programs are marked with a "*")
> 
> Language features:

[snip]

> - Separation between read-only strings (type string) and read-write byte
>   sequences (type bytes). Activated by command-line option -safe-string.
>  (Damien Doligez)

PR #185, opened 2017-11-15
check more carefully addition in sub, fixes #182

### LINKED ISSUE #182: Out-of-bounds error with sub
It's another arithmetic overflow issue.  Here's a test case with a 32-bit OCaml:

```ocaml
(* bounds.ml *)
open Cstruct
let x = create max_int
let y = sub x (max_int - 1000) 1
let z = sub y max_int max_int
let () = blit_from_bytes (Bytes.create 100) 0 z 0 100
```

```
$ ocamlfind opt -package cstruct -linkpkg bounds.ml
$ ./a.out 
Segmentation fault
```

hannesm:
there are a lot of `check_bounds t (off+len)` in `cstruct.ml`, I wonder whether we need to guard that addition for overflows (if so, https://github.com/ocaml/ocaml/pull/1060 would be handy, but this is 4.05+)

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -125,7 +125,8 @@ let debug t =
 
 let sub t off0 len =
   let off = t.off + off0 in
-  if off0 < 0 || len < 0 || not (check_bounds t (off+len)) then err_sub t off0 len
+  let offlen = off + len in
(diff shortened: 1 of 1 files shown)


djs55:
lgtm

PR #186, opened 2017-11-17
test 4.06

hannesm:


djs55:
Looks good. I'll be interested to see if OCaml 4.06.0 fails -- when I tried building it locally I hit a deprecated function, possible workaround in #187

PR #187, opened 2017-11-17
Fix the build on OCaml 4.06.0

djs55:
The `Bigarray.Array1.map_file` function is deprecated and we should use
`Unix.map_file` in future, however `Unix.map_file` is new in OCaml 4.06
and we still support OCaml 4.04.2.

This patch removes the deprecated warning in the `unix/` directory
as a work-around.

Signed-off-by: David Scott <dave@recoil.org>

### DIFF
--- a/unix/jbuild
+++ b/unix/jbuild
@@ -2,4 +2,7 @@
  ((name cstruct_unix)
   (wrapped false)
   (public_name cstruct-unix)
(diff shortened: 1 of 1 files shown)


talex5:
BTW, you can disable warnings for individual statements now - see https://caml.inria.fr/mantis/view.php?id=6714

PR #189, opened 2017-11-17
No base bytes

hannesm:
cc @djs in reply to https://github.com/mirage/ocaml-cstruct/pull/183#issuecomment-345259444 this looks safe


djs55:
Looks safe to me too.

PR #190, opened 2017-11-17
Add workspace file for testing against multiple versions

rgrinberg:
make all-supported-ocaml-versions build & tests against 4.03.0, 4.04.2, 4.05.0, 4.06.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,6 @@ pkg-%:
 PKGS=$(basename $(wildcard *.opam))
 opam-pkg:
 	$(MAKE) $(PKGS:%=pkg-%)
+
+all-supported-ocaml-versions:
+	jbuilder build --dev @install @runtest --workspace jbuild-workspace.dev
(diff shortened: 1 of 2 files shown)


hannesm:
something similar used to be in the `lwt` repository (where then `jbuilder build` always failed for me).  since different people name their switches differently, IMHO the workspaces do not belong into version control.

rgrinberg:
Note that this wouldn't cause any failures outside of using this target. The issue with hard coding switch names is a bit annoying of course, I'd be interested in ways of improving it if possible.

hannesm:
ok, then it was differently setup in lwt.  nevertheless, I don't think having this (completely unused, since our CI systems work differently -- there is one job per OCaml version) in the repository is a good idea.

rgrinberg:
There are many targets in the Makefile that are unused by the CI. Nevertheless, if the the maintainers don't end up using this target then of course it will be redundant and unmaintained, and eventually harmful. So yeah, if there's no interest from the other maintainers then it's best to close this.

samoht:
Yea, I am not sure about that workspace thing as well. Closing (but could be happy to re-open later if more maintainers are interested).

PR #192, opened 2017-11-21
Wrapped false everywhere

rgrinberg:
This is omitted in a couple of libraries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all:
 	jbuilder build @install --dev
 
 clean:
-	rm -rf _build *.install
+	jbuilder clean
 
 test:
 	jbuilder runtest --dev
--- a/lib/jbuild
+++ b/lib/jbuild
@@ -2,6 +2,7 @@
   ((name cstruct)
    (public_name cstruct)
    (modules (cstruct))
+   (wrapped false)
    (libraries (sexplib))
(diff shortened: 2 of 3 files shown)


hannesm:
could you please remind me what `wrapped false` does exactly?  I thought it is only relevant in case you have libraries which span over multiple modules (which is not the case for `lib/` - which only includes a single module).  thanks!

rgrinberg:
You're essentially right. For libraries with a single module, `(wrapped false)` changes nothing. However, if you don't intend to use module aliases, I think making that explicit with `(wrapped false)` is the right thing to do. Since relatively few people know that it's possible to define your "top level" alias module to carefully control what gets exposed, they're usually surprised with the behaviour when a new module is added to a single module library without `(wrapped false)`.

hannesm:
@rgrinberg thanks for the explanation.  I compared a build of `lib/` with and without `wrapped false` in jbuild, and it leads to the same binaries.  I don't have much opinion on the "right thing" in here, mainly because I don't see any plans to extend cstruct with more modules (netiher to split the existing one).  I'll others decide on this.

rgrinberg:
Just a note, some sub libraries have `(wrapped false)` set already (see cstruct-lwt) for example. So I think that we should decide one way or the other.

PR #193, opened 2017-12-07
add support for native-endian integers

vbmithr:


hannesm:
I'm sceptical since I never had a use case for native-endian integers.  Is there a specific protocol / wire format / disk format which requires native-endian (by definition this won't work as a communication protocol between two computers with different endianness, nor exchanging data between such application running on systems with different endianess)!?

avsm:
I agree; it would be good to understand the usecase here.

vbmithr:
I'm not sure why I needed that, but I know I don't use it anymore. Closing this for now.

PR #201, opened 2018-06-26
Do not depend explicitly on deprecated ppx_driver

edwintorok:
To avoid having to pull in a deprecated package depend on either
`ppx_driver` or `ppxlib`, apparently the only reason we depend on it is
to ensure the version is new enough, ppx_cstruct does not require it
directly.

I kept both ppx_driver and ppxlib so that the dependency works with
older compiler versions too.

Do not put version constraints on `ppxlib`, any version is newer than
`ppx_driver`, and if the version constraint is not satisfied then opam
would fall back installing a new `ppx_driver` instead of upgrading an
existing `ppxlib`.

Tested with opam 2.0.0~rc3 and OCaml 4.06.1

### DIFF
--- a/ppx_cstruct.opam
+++ b/ppx_cstruct.opam
@@ -21,7 +21,7 @@ depends: [
   "ounit" {test}
   "ppx_tools_versioned" {>="5.0.1"}
   "ocaml-migrate-parsetree"
-  "ppx_driver"    {test & >= "v0.9.0"}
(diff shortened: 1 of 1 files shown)


edwintorok:
The alpine build failure is https://github.com/janestreet/core/issues/109, the fedora one is an external solver error. (not sure what can be done about it other than starting to use opam2 with the new solver)

PR #204, opened 2018-08-26
avoid compiler-libs runtime dependency in ppx_cstruct

hannesm:
What I observed while demoing on stage at bornhack is a `hello.ukvm` which is rather large (>10MB), so I wondered what is in there.  Turns out, the OCaml compiler libs once again managed to find their way into all the unikernels.

After figuring out (`ocamlfind query -r <my-package>`) where this originates from, I stumbled upon `mirage-profile` which (run-)depends on `ppx_tools_versioned`.  The culprit is `ppx_cstruct`, quoting their META file:
```
# This line makes things transparent for people mixing preprocessors
# and normal dependencies
requires(-ppx_driver) = "cstruct ppx_tools_versioned"
```

I looked into other `ppx_*`  META files and discovered that the line above should only list run-time dependencies, which is here not the case.  I permutated the `ppx/jbuild` file in various combinations, and the one in this PR leads to a working `ppx_cstruct`, which does not have a run-dependency on `ppx_tools_versioned`.  Since I'm no ppx neither jbuild/dune expert, I'd appreciate comments whether this is the right way to solve this issue.

My environment: I tested OCaml 4.06.1 and 4.07.0 (on FreeBSD/amd64), using jbuilder 1.0+beta20, also dune 1.0.1 and 1.1.1 (all three of them show the same behaviour).

With this patch, `hello.ukvm` is back to 2.6MB.  A merge and release would be highly appreciated @mirage/core //cc @rgrinberg

### DIFF
--- a/ppx/jbuild
+++ b/ppx/jbuild
@@ -9,6 +9,4 @@
   (preprocess (pps (ppx_tools_versioned.metaquot_404)))
   (libraries
    (ocaml-migrate-parsetree
-    ppx_tools_versioned
-    ppx_tools_versioned.metaquot_404
     bigarray))))


samoht:
Which versions of the packages are you using?

I think the issue was supposed to be fixed by https://github.com/ocaml-ppx/ppx_tools_versioned/pull/6

hannesm:
@samoht ah, interesting.  the PR of ppx_tools_versioned you mention is not part of any released version :/.

and it turns out that you're correct: ppx_tools_versioned 5.1 does _not_ lead to huge binaries, but 5.2 does.  so instead of merging this PR, a release of ppx_tools_versioned would be more appropriate?

hannesm:
I asked in https://github.com/ocaml-ppx/ppx_tools_versioned/pull/6 for a release, and after testing with ppx_tools_versioned pinned to master, the released ppx_cstruct 3.2.1 does not include the above mentioned runtime dependencies.

So, this PR is not needed, but rather a release of ppx_tools_versioned.

avsm:
should we also put a conflict on ppx_tools_versioned 5.2 to avoid it being selected?

hannesm:
@avsm I'd appreciate (a) ppx_tools_versioned 5.2.1 release and (b) `available: false` in ppx_tools_versioned 5.2 (or removal thereof) //cc @let-def

avsm:
I did a quick point release to save @let-def the trouble... https://github.com/ocaml/opam-repository/pull/12519

Issue #200, opened 2018-01-20
of_sub_string etc


cfcs:
We currently have these functions for initializing a new Cstruct:
```ocaml
  val Cstruct.of_bigarray ?off:int -> ?len:int -> buffer -> t
```
and
```ocaml
  val Cstruct.of_string ?allocator:(int -> t) -> string -> t
  val Cstruct.of_bytes ?allocator:(int -> t) -> bytes -> t
  val Cstruct.of_hex string -> t
```
It would be nice, and would seem more consistent, if it would be possible to construct a `Cstruct.t` directly from a slice of a `string`/`bytes` (like you can with `of_bigarray`) instead of having to first do a `String.sub`/`Bytes.sub` (which means a redundant memcpy operation) before the copy into the new `Cstruct.t`.

Also, why is there no `?allocator` for `of_bigarray`? Is that because it doesn't actually copy, but maps to the same memory? If so, that should probably be documented?

XVilka:
@cfcs slicing was added, can this one be closed then?

cfcs:
Yes, thanks!

cfcs:
(I guess the `?allocator` question is still relevant, but I have forgotten why I cared when I reported this)

PR #208, opened 2018-08-30
Add offset and length for of_string/of_bytes

XVilka:
Added for improving parsing performance, reducing the need of string/bytes copy.
See https://github.com/mirage/ocaml-cstruct/issues/200


avsm:
This also lgtm

lib/cstruct.mli
```diff
@@ -214,13 +214,15 @@ val create_unsafe : int -> t
     to leak sensitive information.
 *)
 
-val of_string: ?allocator:(int -> t) -> string -> t
-(** [of_string ~allocator str] is the cstruct representation of [str],
+val of_string: ?allocator:(int -> t) -> ?off:int -> ?len:int -> string -> t
+(** [of_string ~allocator ~off ~len str] is the cstruct representation of [str]
+    slice located at [off] offset and of [len] length,
```
avsm: Not directly relevant to this PR, but it may be worth explicitly clarifying that the memory is copied here, and there is no sharing between the fresh cstruct and the string or bytes.

XVilka:
It is green, failures seems related only to the network connectivity.

Issue #206, opened 2018-08-30
Support bi-endian struct definitions


XVilka:
Currently Cstruct allows to declare either big-endian or little-endian structures with corresponding getters and setters. Would be really handy to support also `[@@bi-endian]` attribute to generate both. Or maybe any better way for structures that can handle both endians.

avsm:
Out of curiosity, what for?

XVilka:
For example in embedded file formats can have the same format structure, but without defined endianess. So program should be able to parse the same structure in both big endian and little endian for different files.

avsm:
That's true -- I believe pcap defines the endianness at the start of the file.

XVilka:
@samoht any suggestions on how to proceed? Because I cannot use the same name as with big-endian/little-endian, I had to provide different function names or wrap them into the module. What is the preferred approach?

samoht:
No strong opinion on the names but I guess using `BE` and `LE` modules like in [ocplib-endian](https://github.com/OCamlPro/ocplib-endian) would make sense.

XVilka:
Fixed in https://github.com/mirage/ocaml-cstruct/commit/554aa364401832c94db5c89ea8ddda195b12eb47

PR #212, opened 2018-09-12
Basic bi-endian support

XVilka:
This PR tries to solve https://github.com/mirage/ocaml-cstruct/issues/206
It is my first time to write PPX, so not sure if I am going into the right direction.


avsm:
Thanks for this! On first glance through the code, it looks sensible to me.

XVilka:
It is green, failures seems related only to the network connectivity.

XVilka:
Added some tests, so if you are fine with this - please merge.

Issue #205, opened 2018-08-30
Bit fields support


XVilka:
It is not uncommon when C structures defined with bitfields
```c
struct {
    unsigned int a : 2;
    unsigned int b : 4;
} some;
```

XVilka:
Note that it would require a different reading/writing logic depending on the endianess:
![image](https://user-images.githubusercontent.com/203261/50396563-b9388c80-07a5-11e9-8282-40069954c555.png)

![image](https://user-images.githubusercontent.com/203261/50395931-ab810800-07a1-11e9-86a1-e86618157d8b.png)
See http://mjfrazer.org/mjfrazer/bitfields/ for more information.

PR #215, opened 2019-01-03
[WIP] Initial bitfields support

XVilka:
Trying to address the problem of supporting bitfields (https://github.com/mirage/ocaml-cstruct/issues/205), please see the initial draft and provide your opinion/criticism/suggestions.

Please note, for now I implemented bitfields only for an assumption that it is GCC-like behavior, with `__attribute__((packed))` or `#pragma pack (push,1)`

At this point only reading is implemented. Once it is polished and ready - I will do writing too, since more complex.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -31,6 +31,7 @@ type buffer = (char, Bigarray_compat.int8_unsigned_elt, Bigarray_compat.c_layout
  * indeed maintains this invariant.
  *)
 
+(* Note that offset is stored in both bits and bytes! *)
 type t = {
   buffer: buffer;
   off   : int;
@@ -256,7 +257,7 @@ let compare t1 t2 =
 let equal t1 t2 = compare t1 t2 = 0
 
 (* Note that this is only safe as long as all [t]s are coherent. *)
-let memset t x = unsafe_fill_bigstring t.buffer t.off t.len x
+let memset t x = unsafe_fill_bigstring t.buffer (t.off/8) t.len x
 
 let create len =
   let t = create_unsafe len in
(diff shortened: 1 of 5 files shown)


XVilka:
Travis error seems unrelated:
```
opam show /tmp/opam-ci-192 --raw
[WARNING] Failed checks on cstruct package definition from source at file:///tmp/opam-ci-192:
             error 57: Synopsis and description must not be both empty
Errors.
             error 57: Synopsis and description must not be both empty
'unset TESTS; OPAMYES=1 export OPAMYES; set -uex; cp cstruct.opam /tmp/opam-ci-192 && opam show /tmp/opam-ci-192 --raw | opam lint -' exited 1. Terminating with 1
The command "bash -ex ./.travis-docker.sh" exited with 1.
```

avsm:
Thanks very much for all the hard work you've put into this!  I'm a little torn about the approach, since we are reimplementing some difficult pieces of a C compiler directly within OCaml.  In particular, the restriction on attributed-packed is what has me worried.

Before I start reviewing this, I wanted to open one alternate possibility.  The [ctypes](https://github.com/ocamllabs/ocaml-ctypes) library has comprehensive infrastructure to do compile-time probing of the C compiler in use (including cross-compilation) in order to determine various struct layouts.  I've wanted a cstruct->ctypes bridge for some time, so one possibility is to use Ctypes' precise definitions of C layout to implement the Cstruct ppx for bitfields.

Another high level question: what is the usecase you want to solve with bitfields? I just want to make sure that the restrictions placed on the current PR (e.g. packed structs only) are compatible with what you need (e.g. is it a bitfield from a network protocol?)

XVilka:
I cleaned up the code, solved the problem with length handling and now tests are green.

ppx/ppx_cstruct.ml
```diff
@@ -254,6 +439,10 @@ let output_hexdump _loc s =
                 [%expr Printf.bprintf _buf "0x%lx\n" ([%e Ast.evar (getter_name s f)] v)]
               |Prim UInt64 ->
                 [%expr Printf.bprintf _buf "0x%Lx\n" ([%e Ast.evar (getter_name s f)] v)]
+              |Bitfield (prim,len) ->
+                [%expr Printf.bprintf _buf
+                    (format_of_bitfield prim len)
```
XVilka: Fixed also hexdump to use valid format strings for bitfields, but not sure how to embed the function in here properly. PPX and AST documentation is really, really bad.

XVilka:
Added also a fix for hexdump, but don't know how to embed the function properly in the PPX.

XVilka:
Wait, it will not work for big endian machines, since on BE it should read bit-by-bit from the most significant bit, not just swap bytes and read from the least significant bits.

XVilka:
Ok, pushed also the big endian proper support, tested with various bitfield layouts.

XVilka:
With big-endian bitfields I met one problem - basically I need to know the size of the structure to count bits from the end, but `t.len` stores the size of the Cstruct buffer instead. Any advice?

XVilka:
So? What do you think about this approach?

lib/cstruct.ml
```diff
@@ -254,41 +255,869 @@ external swap64 : int64 -> int64 = "%bswap_int64"
...
+                    else
+                        (* For now just throw an exception *)
+                        err_invalid_bounds "get_bits_uint8" t i 2
+                in
+                (raw land mask) lsr bithead
+
+(* FIXME: Do not tie up itself to the byte as granularity unit *)
+(* This function cannot read more than 16 bits *)
+let get_bits_uint16_le swap p t (o:int) (i:int) =
+    Printf.printf "get_bits_uint16 [%x] bits @ %x\n" i o;
```
dinosaure: Should be removed?

XVilka: Yes, of course I remove this, once it is more or less ready. Along with comments cleanup. I want more general concept review first, so I know where to change/fix/etc.

lib/cstruct.ml
```diff
@@ -254,41 +255,869 @@ external swap64 : int64 -> int64 = "%bswap_int64"
...
+    else
+        get_uint64 swap p t i
+
+
+(* It also should always start from the most significant bit for big endian and least significant bit on little endian machine
+ * Basically on big endian machine we should read them in reversed order, bit by bit *)
+(* Currently it always starts from the least significant bit
+ * How to handle the byte-peeking then? *)
+(* This function cannot read more than 8 bits *)
+let get_bits_uint8_le swap p t o i =
```
dinosaure: Hmmhmm, it seems that the predicate on this function is to be able to access at, at most, 2 bytes when an asked series of bits can overlap 2 bytes (like, give me 8 bits when `off = 4`). I think, this predicate adds a big complexity when the _at most_ can not be assert at the beginning - and this why, if I understand correctly, we need `get_uint16_chop_boundary`.

I think we can split this function (for instance, I'm focus on this function), to little functions here. A safe one which did not need `get_uint16_chop_boundary` (easy case) and the other complex case. It's much more about readability.

XVilka:
Sorry, was busy with other things, will finish the PR in a few days.

XVilka:
Ok, rebased, updated a bit, now it passes tests also. Let me refactor and cleanup a bit, also I think https://github.com/mirage/ocaml-cstruct/pull/231 should be merged first, so I will rebase on top of it.

XVilka:
@dinosaure @avsm can you please take a look again? I resolved your comment on splitting the function, it is rebased upon latest master and passes `dune runtest`.

XVilka:
@emillon you might be interested in reviewing this one too, since you did a lot of changes lately.

lib/cstruct.ml
```diff
@@ -27,10 +27,11 @@ type buffer = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Arr
  * indeed maintains this invariant.
  *)
 
+(* Note that offset is stored in both bits and bytes! *)
```
emillon: Feel free to rename the field as off_bits or encapsulate this in an abstract size type - that's better than a comment in my mind

lib/cstruct.ml
```diff
@@ -86,7 +87,7 @@ let of_bigarray ?(off=0) ?len buffer =
     | None     -> dim - off
     | Some len -> len in
   if off < 0 || len < 0 || off + len < 0 || off + len > dim then err_of_bigarray off len
-  else { buffer; off; len }
+  else { buffer; len; off }
```
emillon: can you remove these hunks from the diff?

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
     let r = ba_get_int64 t.buffer (t.off+i) in
     if swap then swap64 r else r [@@inline]
 
+(* o - bitoffset, i - count of bits *)
+(* It assumes we use __attribute__((packed)) with alignment = 1 byte *)
+(* Here we also assume there is no trailing padding at the end *)
+(* The offset and alignment rules depend on the "mother" type *)
+let align = 8 (* bits, not bytes *)
```
emillon:
```suggestion
let align_bits = 8 (* bits, not bytes *)
```
(or use an abstract type as suggested above)

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
     let r = ba_get_int64 t.buffer (t.off+i) in
     if swap then swap64 r else r [@@inline]
 
+(* o - bitoffset, i - count of bits *)
+(* It assumes we use __attribute__((packed)) with alignment = 1 byte *)
```
emillon: Does this assumption hold in general? What are the alternatives? Should this be configurable? :slightly_smiling_face:

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
...
+(* These functions do not cross the boundary, and just add zeroes *)
+let get_uint16_chop_boundary swap p t i =
+    if i + 2 > t.len then
+        let part = get_uint8 t i in
+        if swap then part lsl 8 else part
+    else
+        get_uint16 swap p t i
+
+let get_uint32_chop_boundary swap p t i =
+    if i + 4 > t.len then
```
emillon: I think that these branches could be refactored in to a `match i - t.len with ...` (or `match () with _ when cond1 -> ... | _ when cond2 -> ...`  which would make all branches at the same level)

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
...
+        get_uint64 swap p t i
+
+let get_bits_uint8_le_boundary_chop swap p t o i =
+  let byteoff = o / align in
+  (* 6 cases total *)
+  (* When they start at granularity boundary *)
+  if o mod align = 0 then
+    (* First case - when they are located near the granularity boundary,
+     * and lesser or equal than a granularity unit (byte) *)
+    (* FIXME: Do not tie up itself to the byte as granularity unit *)
```
emillon: I think it's asking you to fix it :)

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
...
+    if i <= align then
+        let byte = get_uint8 t byteoff in
+        let mask = (1 lsl i) - 1 in
+        byte land mask
+    (* Second case - when they are located near the granularity boundary,
+     * and bigger than a granularity unit (byte) *)
+    else
+        (* Check if it crosses the boundary *)
+        if i mod align = 0 then
+            (* TODO: Consider also the length of the buffer here *)
```
emillon: I think it's asking you to do this :)

ppx/ppx_cstruct.ml
```diff
@@ -157,21 +181,55 @@ let op_evar s op = Ast.evar (op_name s op)
...
+    [%expr fun src -> Cstruct.sub src [%e num off] [%e num len]]
+  |Bitfield (prim, len) ->
+    (* It depends on the actual bits count, so if the bits count
+     * lesser than primary type size - we still take primary type,
+     * but if it is bigger - we take the bigger unit then *)
+    if len > 0 && len <= 8 then
+    [%expr fun v ->
+          [%e match prim with
+              |Char |UInt8 ->
+                [%expr [%e m "get_bits_uint8"] v [%e num f.off] [%e num len]]
```
emillon: These bits can be factored out by computing the getter function for each prim type.

lib/cstruct.ml
XVilka: I could make it configurable, then it would become not +50%, but +500% of additional code.

XVilka:
> If this is possible, I suggest building a separate library on top of cstruct that deals with bitfields and unaligned accesses. If this would rely on primitives that don't exist yet in cstruct, we can add these more easily.

I will think about it. Might be a good idea, something like `cstruct-bitfield`, adding the ability to parse `[@bits X]` attribute and expanding it before passing the rest to Cstruct. I only have to think about on how to separate the code from the Cstruct.

lib/cstruct.ml
```diff
@@ -27,10 +27,11 @@ type buffer = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Arr
  * indeed maintains this invariant.
  *)
 
+(* Note that offset is stored in both bits and bytes! *)
 type t = {
   buffer: buffer;
-  off   : int;
   len   : int;
+  off   : int;
```
dinosaure: Same as @emillon, can we remove this diff?

lib/cstruct.ml
```diff
@@ -282,13 +283,923 @@ let get_uint64 swap p t i =
...
+  (* 6 cases total *)
+  (* When they start at granularity boundary *)
+  if o' mod align = 0 then
+    (* First case - when they are located near the granularity boundary,
+     * and lesser or equal than a granularity unit (byte) *)
+    if i <= align then
+        let short = get_uint16_chop_boundary swap p t byteoff in
+        let mask = (((1 lsl i) - 1) lsl o') lsl bithead in
+        (* Printf.printf "bithead = %d, i = %d, o = %d, o' = %d\n" bithead i o o';
+        Printf.printf "1st case - 0x%x & 0x%x\n" short mask; *)
```
dinosaure: Debugging? You should delete this comment.

XVilka:
I skimmed through code and I don't think it is feasible to do it, since it affects other, non-bit fields, so I have no idea how to transform the AST the way that will be still useful for Cstruct and it will use right offsets, affected by alignments and bits count.

lib/cstruct.ml
dinosaure: At the end, this update adds lot of headache when `{get,set}_uint{8,16,32,64}` did not changes and believed on `off` concerns bytes (and not bits).

For example, `memset` divides by 8 `off`, I don't understand why.

And it seems that any `get_bits` requires an absolute `off` as an argument (and did not rely on internal `off` offset).

XVilka: I can change these functions to use `t.off` if you think it is better, that's not a problem.

lib/cstruct.ml
avsm: The only viable alternative if we do not assume this is to switch to using ocaml-ctypes (which does handle real C alignments) in order to determine a struct layout (see [my earlier comment](https://github.com/mirage/ocaml-cstruct/pull/215#issuecomment-451433761)).

I'm in two minds about how to handle this usecase -- an alternative approach to solve the problem of bitfields is to define a ctypes struct layout, do the stub generation/probing that ctypes does already, and then re-expose those offsets using the Cstruct conventions.  That would involve actually removing almost all the code in ppx_cstruct since the C compiler would do the bulk of the work then :-)

avsm:
I am tending to agree with the review comments here; the complexity of bitfields is enormous _vs_ the general trend for cstruct to be more explicit with respect to atomicity of multi-word operations.

Looking at the [squashfs example](https://github.com/plougher/squashfs-tools/blob/master/squashfs-tools/squashfs_compat.h#L440) again, it strikes me that this would be a straightforward ctypes binding since the struct/unions are all there already, iff [bitfields were added](https://github.com/ocamllabs/ocaml-ctypes/issues/226) to ctypes. @XVilka, it might be worth giving that a try -- if the non-bitfield pieces of SquashFS are straightforward to bind with ctypes, then the stub generation support will give really good compiler-driven layout information and so work with both packed and non-packed attributes. (cc @yallop)

XVilka:
Uh, I am not sure... I already have quite a lot of code written with
Cstruct. Switching to Ctypes stubs can take a lot of work with zero
benefits. I might want to consider this in the future, but not right now. I
can just keep to use my Ctypes fork, just wanted to push it mainstream... I
also can do an extension over Cstruct, but have no idea how to do that to
be honest.

avsm:
I've been thinking for some years that the Ctypes description should be the 'type' behind the Cstruct definition.  Right now the ppx just interprets a syntax directly, but doesn't really have an OCaml data structure that describes it.

The rough idea is to use Ctypes 'behind the scenes' in ppx_cstruct in order to determine the layout/offsets of each value, and then expose the Cstruct offsets as today.  The main difference is that Ctypes will give accurate offsets for any C compiler, _vs_ the reimplementation we have to do today in ppx_cstruct.

avsm:
Another thought with this patchset is that to reduce the risks to non-bitfield users, we could also copy the code into a ppx_cstruct_bitfield package here, and maintain it as an experimental ppx alongside ppx_cstruct.

If we advertise it as experimental, it's ok for it to not keep up with the featureset in ppx_cstruct (which I doubt will develop very fast anyway), and @XVilka doesn't need to keep a private fork for his usecases.

XVilka:
Ok, I will think how to proceed.

XVilka:
Should I close this one? Seems unlikely it will be the part of Cstruct itself. I just wonder if it is possible to do somehow _outside_ of Cstruct without much of the copy-pasting. Any advice would be welcomed.

avsm:
I never got a reply to my comment earlier [proposing a way forward](https://github.com/mirage/ocaml-cstruct/pull/215#issuecomment-474840914).

XVilka:
@avsm oh, sorry. I indeed can make a separate module/package for ppx. I will do this.

mrvn:
Bitfields should not assume or require attribute packed. I'm not even sure packed has the effect you assume. Lets start with how bitfields are layed out by the compiler. Hopefully from the examples you can build some tests to cover all the corner cases:

The layout of a bitfield can be complex. The first problem is that you have to figure out if the bitfield starts counting from the MSB or LSB. It usually matches the endianness but you just have to know what the compiler will do. I don't believe there is any define for this. All the examples below match what gcc does. I believe this is universal but that's what test cases are for.

The next thing is that the compiler will pack members of a bitfield into the specified type one after the other but a member never bridges the border of the specified type. So for example

    struct {
        uint8_t a:5;
        uint8_t b:5;
        uint8_t c:5;
    };

will be 3 bytes, each holding 5 bit data and 3 bit padding. The bits are not packed into 15 bit data and 1 bit padding as you get with

    struct {
        uint16_t a:5;
        uint16_t b:5;
        uint16_t c:5;
    };

Alignment and padding rules of the bitfield follow the specified type. This can be tricky when you mix types:

    struct {
        uint16_t a : 7;
        uint8_t  b : 5;
    };

Here a takes the first 7 bits, obviously. But then there is 1 padding bit, 5 bits for b and 3 padding bits again. b is packed together with a into a single uint16_t but aligned so it doesn't cross the uint8_t border.

    struct {
        uint8_t  a : 5;
        uint16_t b : 7;
    };

Similar a and b are combined into a single uint16_t. But now there is no padding between them as b does not cross an uint16_t boundary.

Let me describe this in another way. Bitfields will be packed as closely as possible except padding is added so that each member can be accessed (as if) by a single aligned read/write of the specified type for that member. So in the last example b must be readable with an aligned 16 bit access and then masking and shifting. No padding is required for this. In the example before that b must be readable with a single 8 bit access, which requires 1 padding bit.

XVilka:
@mrvn thanks a lot for the insights. Sadly, after multiple tries I decided the resulting code would be too complex. Thus, I am not planning to pursue this goal anymore - I handled all bitfields manually in the corresponding parsers.

XVilka:
I am closing this then.

PR #217, opened 2019-01-07
Update opam files for 2.0 format

XVilka:
Since now opam 1.2 is dead, time to update. Basically copied opam files from opam-repository and changed URL. It should fix complains from Travis too.

### DIFF
--- a/cstruct.opam
+++ b/cstruct.opam
@@ -1,22 +1,29 @@
-opam-version: "1.2"
+opam-version: "2.0"
 maintainer:   "anil@recoil.org"
 authors:      ["Anil Madhavapeddy" "Richard Mortier" "Thomas Gazagnaire"
                "Pierre Chambart" "David Kaloper" "Jeremy Yallop" "David Scott"
                "Mindy Preston" "Thomas Leonard" ]
 homepage:     "https://github.com/mirage/ocaml-cstruct"
 license:      "ISC"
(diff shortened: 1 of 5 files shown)


XVilka:
Hmm, should we add also alcotest to the dependencies?
```
- File "lib_test/jbuild", line 4, characters 22-30:
- 4 |  ((libraries (cstruct alcotest))
-                           ^^^^^^^^
- Error: Library "alcotest" not found.
- Hint: try: dune external-lib-deps --missing -p cstruct @runtest
```

XVilka:
Fixed the alcotest problem, but I don't get why Alpine jobs are failing.

avsm:
thanks very much; i'll fix the CI in a follow up PR

cstruct.opam
```diff
@@ -1,22 +1,29 @@
...
 ]
-build-test: ["jbuilder" "runtest" "-p" name]
-
 depends: [
-  "jbuilder" {build & >="1.0+beta10"}
-  "sexplib"
+  "ocaml" {>= "4.03.0"}
+  "jbuilder" {build & >= "1.0+beta10"}
+  "sexplib" {< "v0.12"}
   "alcotest" {test}
```
avsm: This was the faulty line btw -- `{test}` should `{with-test}` in opam2. `opam package-upgrade *.opam` should automate this

PR #219, opened 2019-02-01
Remove trailing spaces in hexdump output

emillon:
Hello,

I added some tests for `hexdump_pp`.

The formatting seems a bit off to me:
- there are trailing spaces
- there is no newline at the end

I believe that the trailing spaces should be removed, but I'm not too sure about the trailing newlines.

Let me know what you think.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -412,13 +412,22 @@ let of_hex str =
   | cs, i, _, _ -> sub cs 0 i
 
 let hexdump_pp fmt t =
+  let before fmt =
+    function
+    | 0 -> ()
+    | 8 -> Format.fprintf fmt "  ";
+    | _ -> Format.fprintf fmt " "
+  in
+  let after fmt =
+    function
+    | 15 -> Format.pp_force_newline fmt ()
+    |  _ -> ()
+  in
   Format.pp_open_box fmt 0 ;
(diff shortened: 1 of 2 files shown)


emillon:
(feature was initially included in #175 where there's some discussion about this)

avsm:
Lgtm; fixing the trailing spaces and newline sounds good!  I just noticed that cstruct doesn't use ocaml-hex from mirage, but the extra dependency probably isnt worth it

hannesm:
@emillon great, thanks!

emillon:
I removed the spaces and used a vbox instead of forcing a newline.

As for the trailing newline, after trying I don't think it's `hexdump_pp`'s responsibility to add a break at the end.

For example, let's say we're trying to write some text and a hanging hexdump:

```
Some text 00 01 03....
          11 ....
          3f ....
```

At the moment there is no trailing newline. But forcing a break will insert spaces (indicated by `␣`):

```
Some text 00 01 03....
          11 ....
          3f ....
␣␣␣␣␣␣␣␣␣␣
```

Which makes it impossible to output things like:

```
Some text 00 01 03....
          11 ....
          3f ....
Something else
```

So, we can merge this as this already fixes the trailing spaces, and try to think of a better solution once it's merged with `hex` :)

avsm:
thanks!

PR #220, opened 2019-02-03
Remove old compatibility packages for `cstruct.lwt`, `cstruct.async`,

avsm:
`cstruct.ppx` and `cstruct.unix`.

These were deprecated in cstruct.3.0.0 in favour of counter part
libraries with a dash in the name (`cstruct-lwt`, `cstruct-async`,
`cstruct.unix`) or `ppx_cstruct` for the PPX extension.

Keeping these around makes it harder to embed in a duniverse, since
old users of the name will result in a dependency cycle.  There
are very few packages still using it though, and this release will
flush them out...


avsm:
I'm going to go ahead and do a quick release with this in order to see what the revdep compatibility damage in opam-repository is.  That's the easiest way to fix all the offending packages.

Issue #226, opened 2019-02-27
ppx tests - do they run?


emillon:
Hi,

I noticed that ppx tests are built, but not executed. They define a main function, but it does not run:

https://github.com/mirage/ocaml-cstruct/blob/b30ec693e1218bc3e1b3333259312f4555300b88/ppx_test/pcap.ml#L173

Is this intended? otherwise I can set up a test harness and a `.expected` file so that we can compare the output.

Let me know what you think. Thanks!

avsm:
The original intention was to make sure they compile — it should do no harm to run them as well indeed :)

PR #227, opened 2019-02-28
Run ppx tests

emillon:
Hi,

To solve #226, the tests are now executed.

@XVilka, the tests you added in #212 were actually failing. This is because it relies on roundtrip through a `uint8_t` field but larger values are passed. I think that the intent was to check that the values are equal modulo 256, but let me know if that does not seem right.

### DIFF
--- a/lib_test/dune
+++ b/lib_test/dune
@@ -1,11 +1,9 @@
-(executables
+(executable
  (libraries cstruct alcotest)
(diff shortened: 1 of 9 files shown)


XVilka:
Right, good catch!

emillon:
Thanks but TBH your test caught it :)

PR #231, opened 2019-03-04
Suppress unused warning on "sizeof" & cenum functions

emillon:
Followup for #228: I missed this one because there was no `.mli` for the test file.

This is a bit more rare in practice, since in many cases users of `ppx_cstruct` add a `let () = assert (sizeof_s = n)`.

(for which we could add first class support maybe?)

### DIFF
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -329,33 +329,33 @@ let output_struct_sig loc s =
             (op_typ op)))
     (ops_for s)
 
+let convert_integer f suffix =
+  (fun i -> Exp.constant (Pconst_integer(f i, suffix))),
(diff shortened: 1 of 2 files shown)


emillon:
I also added suppression attributes on the cenum & sexp functions - that can go in a separate PR if you'd prefer.

XVilka:
Looks good, but not sure about `let_` name.

emillon:
Indeed, that could be confusing! I renamed to `declare`.

avsm:
Darn, this conflicts against `master` now

emillon:
@avsm I just rebased - and did the same treatment I did in #233, namely go through an intermediate `enum_op` type to deduplicate the logic for emitting structures and signatures.

avsm:
thanks, it's easier to read now with the intermediate step

Issue #230, opened 2019-03-04
ppx: do not emit code for unused fields


emillon:
Hi,

Sometimes, certain structure fields are not used, in particular for padding:

```ocaml
  [%%cstruct
  type segment = {
    gref: uint32_t;
    first_sector: uint8_t;
    last_sector: uint8_t;
    _padding: uint16_t;
  } [@@little_endian]
  ]
```

(from [mirage-block-xen](https://github.com/mirage/mirage-block-xen/blob/26b19f3f2be01c505687b3cec1666f3d4f5d73a7/lib/blkproto.ml#L303-L310))

We could detect that (if the field start with a `_`) and not emit accessor functions for this field. This can save some space in the generated code until the compiler is able to remove it.

Or is this too magical?

(in the particular case of padding, an alternative is to add a `[@pad_to 64]` attribute, but it seems a bit too specific)

XVilka:
I agree this is a good idea. I use Cstruct a lot for parsing file formats and protocols, and it will save some code, since it is quite often you don't need every single member in the header/some entry.

XVilka:
In the future it might be good to add an alignment attribute though, thus completely eliminating padding members.

avsm:
Skipping underscore fields like this seems fine to me. Could also support just “_” to signify an unbound field.

emillon:
OK! I'll prepare a PR.

PR #233, opened 2019-03-07
Ignore underscore fields

emillon:
This adds special handling to fields whose name start with an underscore: field converters are not generated for this field (#230). This can save some bytes in the binary, but more importantly this conveys intent a bit better.

This is technically a breaking change, but I don't expect that this was used in the wild.

### DIFF
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -186,25 +186,6 @@ let type_of_int_field = function
   |UInt32 -> [%type: Cstruct.uint32]
   |UInt64 -> [%type: Cstruct.uint64]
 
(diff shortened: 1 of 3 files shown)


emillon:
This includes a pretty large refactor (first list a list of ops for a structure, then emit signatures or structures only based on these ops), which can be done in its own PR if necessary.

emillon:
`{_ : t}` is not valid syntax, so we cannot use this for "ignore it and don't name it".

emillon:
Last question is: do we want to print ignored values in hexdumps?

I think that we do not, and the implementation is simpler, so I removed them.

avsm:
Thanks!

Issue #232, opened 2019-03-07
Let's use ocamlformat


emillon:
Hi,

[ocamlformat](https://github.com/ocaml-ppx/ocamlformat) is getting pretty mature, and so far it's been a good experience for the new mirage libraries.

I suggest that we build on that traction and start converting existing libraries. Tools like [merge-fmt](https://github.com/hhugo/merge-fmt) will help for existing PRs.

If there's interest, I'll try to write a configuration that's close to the existing conventions.

Thanks!

samoht:
That would be great!

avsm:
Rather than keeping the existing conventions, I'd prefer to use the standard format in the forthcoming ocamlformat 0.9 release...

emillon:
OK, we can try that.

samoht:
@avsm or we can pick the "right" conventions and try to change the upstream default in 1.0 :-)

PR #234, opened 2019-03-07
Setup ocamlformat

emillon:
- [ ] CI hook
- [x] decide 0.8 / 0.9


emillon:
@avsm here is how it looks with the default configuration (I'm personally not too found but it's better than anything not enforced).

There are not many changes between 0.8 and 0.9 (I made a separate commit), so I say we can go with the 0.8 one and the upgrade will be simple later on.

emillon:
(see #234)

avsm:
Looks fine to me with the default config and 0.8. Will hold off on merge until some of the outstanding PRs are merged

.ocamlformat
```diff
@@ -0,0 +1 @@
+version=0.9
```
avsm: Is this file necessary if we just use the defaults?

emillon:
@avsm I updated this to use the new `conventional` profile.

hannesm:
hmm. First of all, I'm really in favour of using ocamlformat for MirageOS projects! But I'm wondering (a) which ocamlformat is ready for being used and (b) can we come up with an ocamlformat configuration everybody is happy with (at least can tolerate), (c) a ocp-indent configuration that agrees with (b), and (d) setup git hooks. To me, the style used here (also the one in mirage/mirage - which I think is even more verbose) is a bit too verbose (e.g. 4 spaces indent after `->` in a match).

If we hurry too fast (i.e. without (d)), we end up in a state as mirage/mirage is now: it used to be ocamlformatted, but the current state is no longer what ocamlformat would do. I'm wondering (couldn't find any statement about this) whether the `sparse` and `compact` profiles are meant to stay in future ocamlformat versions the way they are.

I suspect (b) will require discussions (or a vote), I'd be happy using `compact` (if we get a guarantee from upstream that it won't change over time -- otherwise version X used by CI will fight with Y used by some developer). Nothing is more annoying (to me at least) having different auto-formatters fighting each other (see https://github.com/ocaml-ppx/ocamlformat/issues/43 - but AFAIK the scope of ocamlformat has been revised, and matching ocp-indent is nowadays in scope).

emillon:
Agree about the CI hook; for now there's a bit of a discussion regarding how to version ocamlformat and how to enable it on CI (https://github.com/ocaml-ppx/ocamlformat/issues/699).

Regarding style, I picked `conventional` based on the discussion in #232 - this new profile is based on feedback on the previous default one which had too many flaws.

samoht:
for what's is worth, I'm using `profile=conventional` and ocamlformat 0.9 in irmin master and I'm pretty happy about it.

hannesm:
@samoht ok, good. what about ocp-indent? or do people not use ocp-indent anymore, and switch over to ocamlformat (is there an emacs hook?)? should ocamlformat profile (version as well?) be part of the repository (some .ocamlformat or sth?) so that others know which style to use (using "default" is not a good idea imho because ocamlformats defaults may be subject to change)?

samoht:
Yes you need to commit a [config file](https://github.com/mirage/irmin/blob/master/.ocamlformat) in the repository, although ocamlformat is also able to read `.ocp-indent` files and do sensible things with it.

Currently I'm using both `ocp-indent` (when I press TAB while editing a file) and `ocamlformat` (I've added a hook to call it on every save) and that works reasonably well. There are plans to the integrated merlin parser directly in ocamlformat to deal with files which cannot parse so the situation will improve and we probably won't need 2 different tools in the longer term.

emillon:
(In case you missed it, there's a `.ocamlformat` file in this PR)

hannesm:
@emillon great

@samoht also great

in my view, still missing: (a) CI hook, (b) decide on a ocamlformat profile, and (c) figure out whether ocamlformat developers will guarantee stability of a profile over releases. From the issue @emillon linked:

> Note that currently the version checking in the config file is actually too lenient. To get reproducible results, it is also required that the version of ocaml used to compile ocamlformat is the same, and perhaps also that of ocaml-migrate-parsetree. This is due to corner cases where the mapping done by ocaml-migrate-parsetree for different compiler versions can lead to slightly different comment placement. More commonly, the parser in different versions does not always give the same source locations, which also affects comment placement. Note that for this sort of version dependency, there isn't a feasible way to test in the formatter code and adjust behavior, it's too late.

which basically means we've to stick with a concrete ocamlformat, ocaml and ocaml-migrate-parsetree version, without being able to specify this in `.ocamlformat` (neither anywhere else)!? this approach is not very convincing to me (I use `git praise` (aka `git blame`) quite a lot do understand why certain lines of code exist, and in which context they were done -- this is certainly harmed by every commit that does "auto-indent", so I'd prefer to have at most a single commit that re-indents with ocamlformat, and then never again (since all subsequent commits enfore the same indentation)), maybe we should wait for a more sustainable story for ocamlformat (but in respect to the comments in their issue tracker, I'm not sure whether that's a goal of ocamlformat)?

from https://discuss.ocaml.org/t/ann-release-of-ocamlformat-0-9 I read that ocamldoc can as well be ocamlformated, which imho we should do -- `parse-docstrings=true` in `.ocamlformat`!?

emillon:
Yes, I wish that it was possible to control versioning more precisely so that formatting only changes when explicitly changing a configuration option (including a compatibility version number). The solution that ocamlformat developers recommend is to always use the latest release, so it will be necessary to fix formatting problems randomly (when a new version is released and picked by the solver, or when the opam images are updated).

I agree that it's not a pleasant developer experience. But it is probably not the endgame for how to enable ocamlformat in a project. One thing to consider is that we (@mirage) are in a good position to test this process on a couple repositories and report that to the ocamlformat team. That might make a good case for explicit versioning.

avsm:
> maybe we should wait for a more sustainable story for ocamlformat (but in respect to the comments in their issue tracker, I'm not sure whether that's a goal of ocamlformat)?

It's a goal of ocamlformat when stable (we're working quite hard on it atm), but has to happen in steps. I'm hoping that we can use a few mirage libraries to beta test some workflows as it heads towards that 1.0 stable release.  Is Cstruct that library? 

@hhugo created https://github.com/hhugo/merge-fmt to help deal with merge conflicts across formatters.  I wonder if something equivalent is possible for `git blame` to normalise away formatting changes as if it were whitespace...

hannesm:
I see your points. Let me try to explain why I'm hesitating with ocamlformat and cstruct.

I really like the idea to have a unified syntax and style across mirage repositories. I'd like to get there with a minimal amount of `git praise` breakage. To achieve this, we should document (maybe [here](https://mirage.io/wiki/packaging)) the process and developer workflow (i.e. list ocamlformat as required tools, maybe as well [here](https://mirage.io/wiki/contributing)). A CI job that fails if the new code does not conform to ocamlformat is required as well to avoid unintentional ocamlformat-breaking merges. To minimize future `git praise` breakages, the update process of ocaml and ocamlformat should as well be explained (i.e. "stick to the latest ocamlformat and latest MirageOS-and-ocamlformat supported ocaml version").

Using mirage as a testbed for ocamlformat is fine with me, and it turns out [this PR](https://github.com/mirage/mirage/pull/935) already introduced ocamlformat for the main mirage repository. Unfortunately, above mentioned documentation and CI is missing, thus the mirage repository is in an ocamlformat-inconsistent state (and there's some confusion (e.g. [this comment](https://github.com/mirage/mirage/pull/978#discussion_r273231782)) about its deployment status) and will (likely) require another `git praise` breaking commit.

The suggestion I have to move forward: test (and deploy CI) with the already-in-betatest mirage repository, and once we're happy with the workflow, migrate other mirage repositories (such as cstruct) to ocamlformat. WDYT?

avsm:
Sounds good to me! And any feedback about life-with-ocamlformat-improvements to the tracker would be most appreciated :-)

avsm:
I'm closing this one for now as it won't be merged as-is. Instead, @emillon @nathanreb and @craigfe are designing an end-to-end workflow that answers [hannes' queries](https://github.com/mirage/ocaml-cstruct/pull/234#issuecomment-484553371) using https://ocurrent.org for this, so that will be proposed in a new PR.

PR #236, opened 2019-03-07
Add inline attributes and use unsafe primitives

Reperator:
This PR adds inline attributes to getters and setters, and uses unsafe primitives. This should be safe since the Cstruct bounds are stricter than the Bigarray bounds. Therefore the Bigarray bounds check will always succeed and can be omitted.

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -277,7 +277,7 @@ let get_uint32 swap p t i =
     if swap then swap32 r else r
 
 let get_uint64 swap p t i =
-  if (i+8) > t.len || i < 0 then err_invalid_bounds (p ^ "uint64") t i 8
+  if (i+8) > t.len || i < 0 then err_invalid_bounds (p ^ ".get_uint64") t i 8
   else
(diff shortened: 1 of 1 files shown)


samoht:
Thanks, that looks great!

/cc @dinosaure

avsm:
This looks good; will merge shortly.

lib/cstruct.ml
```diff
@@ -75,7 +75,7 @@ let err_blit_to_bytes_dst src dst=
   err "Cstruct.blit_to_bytes src=%a dst=%a dst-off=%d len=%d"
     pp_t src bytes_t dst
 let err_invalid_bounds f =
-  err "invalid bounds in Cstruct.%s %a off=%d len=%d" f pp_t
+  err "invalid bounds in Cstruct.%s %a off=%d len=%d" f pp_t [@@inline never]
```
avsm: good idea with this never inline!

Reperator: all credit goes to @chambart

PR #237, opened 2019-03-08
First draft to add capabilities on cstruct

dinosaure:
Let's talk about capabilities under the sun with beers :+1:


dinosaure:
This is a big change, we need to upstream changes on the new `cstruct_core.ml` file but we can talk about the idea under this PR first.

avsm:
interesting use of object types here! I particularly like the `'a align`.

cc @stedolan as he was interested in interfaces for the new effects IO library

hannesm:
Thanks for opening this PR. While working on the network API, and protocol parsers, I really like the idea to have capabilities/properties of buffers:
- read-only - for parsing: a received buffer should be marked read-only, and passed on further
- write-only - not yet sure whether this is really useful - I fail to find a use-case -- checksum computations (which require to read sth) on IP/TCP/UDP get in the way of marking a buffer write-only
- copy-on-reference - something we discovered in the ixy work and in mirage-net-xen recently: here's a buffer, passed down the stack, but don't hold any reference to it (it's owned by someone else and will be zeroed out / reused at a later point)
  - here @gasche outlined an approach to pass a buffer and abstract token (e.g. timestamp), keep the token in the buffer itself as well, and check on every access that the token matches -- if someone preserved a reference with an old token, attempting to access a buffer will lead to an error
  - look at the mirage-net API atm: in `write`, a buffer is allocated by the NIC, and filled by the network stack. any malicious stack could keep references to the buffer and read it later (see https://github.com/mirage/mirage-net-xen/pull/83#pullrequestreview-215769445) (/cc @talex5)

now, reading up on [related](https://github.com/ocaml/ocaml/pull/1961#issuecomment-413468594) [discussions](https://github.com/mirage/mirage/issues/881), as well as face-to-face discussions, I'm wondering:
- since OCaml 4.06.0, the `string` type is immutable
- since [OCaml 4.08.0, reading & writing numbers is supported](https://github.com/ocaml/ocaml/pull/1864)
- the copy-on-reference seems to be complex (and computationally expensive), or require a linear type system -- which we don't have, a conservative solution is to not hand out references to shared, reused memory, and instead do a copy (solo5 does this e.g.)
- memory pressure in respect to bigarrays (see [this comment](https://github.com/ocaml/ocaml/pull/1961#issuecomment-413468594))

out of these reasons, I'm curious what you think about the following proposal:
- get rid of cstruct, replace with bytes for sending / string (unfortunately [this PR](https://github.com/ocaml/ocaml/pull/1864) got merged into 4.08.0 (without `get` functions for `string`, but we can easily develop a library that provides these functions) for receiving data?
- at the OS boundary (i.e. mirage-net-xen), allocate bytes, and blit them after `fill` into the shared memory region --> less brittle, the stack will never see a reference to the shared memory
- for C-FFI, as long as external is marked with `noalloc`, strings/bytes are passed directly (and are not copied to the external heap) AFAIK (please correct me if I'm wrong)

I'm really interested to hear concerns if any, and will attempt to develop a patch to cstruct which uses bytes instead of bigarray -- and see what breaks, and how/whether this improves/degrades performance (any suggestion what an appropriate test would be, otherwise I'd use iperf!?, maybe over TLS (which allocates lots of small cstruct)). I remember having talked about that topic with @samoht who suggested as well to develop a cstruct-backed-by-string/bytes back then, but AFAIK nobody did such an experiment (if someone did, please tell).

talex5:
Another thing that would be useful (given the lack of linear types in OCaml) is a way to revoke a bigarray, so that all further attempts to use cstructs based on it fail at runtime. I made a PR for this a few years ago (https://github.com/ocaml/ocaml/pull/389) but it seems that it would prevent some compiler optimisations.

@gasche wrote there:

> Another route to take would be to declare Bigarray unfit to represent raw memory, and develop an alternative library to do what people have been doing with Bigarrays instead. This alternative should probably be in stdlib for people to actually accept to move to it. But if we have no one willing to do the work of organizing this (which would have a fairly invasive impact on a lot of third-party libraries), I personally think that giving up on the never-freed guarantee is the best move.

If we had a way to revoke a bigarray, we'd probably also want a capability that allowed the holder of a cstruct to get access to the underlying bigarray, so that e.g. mirage-net-xen could prevent its users from extracting the bigarray and taking a new reference to it. Currently, anyone with a cstruct can do that.

avsm:
@talex5 wrote:
> Another thing that would be useful (given the lack of linear types in OCaml) 

My understanding is that it's possible to use GADTs to enforce some linearity, but at the cost of making the API difficult to use since it 'taints' all functions that use it since they have to unpack the GADT to get at any state.

However, one observation I have is that linearity is normally done the wrong way around -- it is embedded in the middle of an application and so hard to enforce.  We could try to do things in reverse: a Mirage scheduler that invokes the application code initialises the linearity state and calls the IO functions.  If a function needs to be non-linear for a while (i.e. weaken the guarantee) then it can do a shift to GC/reference counting, and then when that is done we move back to a linearly tracked dataflow.

The advantage of this model is that for a network stack, we should never need the non-linear access until we get to the application layer.  While linearity is present, is it easy to have a non-copying slicing API.  At the application layer, we can take a decision on how to shift to non-linearity -- e.g. by copying to bytes/string (as suggested by @hannesm above), or doing fancy paging tricks.  I believe this model would also satisfy @talex5's suggestion of bigarray lifetime management since the outer layer would be tracking lifetimes of IO frames explicitly (from the xen driver up).

This is quite a big change, but I believe that putting linearity in the right place in the IO stack is what's missing to have a reasonable lifetime tracking model in the current setup.

hannesm:
@talex5 ah, thanks for that pointer. my TL;DR there is: we should develop a "safe" library that contains a `free`, evaluate it, and if it meets our demands, upstream it into the OCaml runtime. I've also heard there's some interest in an immutable bigarray by some compiler folks. My motivation to go this way is not too high, since I suspect that moving to String/Bytes will give us the same benefits. This shouldn't prevent anyone else from working on such a PR, and evaluating it.

@avsm writes:
> However, one observation I have is that linearity is normally done the wrong way around -- it is embedded in the middle of an application and so hard to enforce. We could try to do things in reverse: a Mirage scheduler that invokes the application code initialises the linearity state and calls the IO functions. If a function needs to be non-linear for a while (i.e. weaken the guarantee) then it can do a shift to GC/reference counting, and then when that is done we move back to a linearly tracked dataflow.

I don't fully understand this proposal:
- What is this MirageOS scheduler? Where does it live? Does it own the memory?
- "If a function needs to be non-linear for a while" <- is this something inferred / automatic, or manual function calls?

TL;DR: IMHO we should for a next cstruct major version have a good story about safety with the different scenarios outlined above. Performance optimization to reduce copying is something we can do in a later step (after evaluating the different proposals). My current intuition is that copying into the OCaml heap for received buffers is fine (and hardly measurable).

Of course, if we can smoothly integrate a linearity invariant, we won't need to copy stuff around. If we fail to do so, better copy than leak information ;)

avsm:
Consider the ownership of a buffer. When it comes in from the outside world, there is precisely one owner (the 'hardware' source it came from).  Then we immediately lose this guarantee by copying it into the OCaml heap.  I'm proposing that rather than passing the single-ownership buffer into a multi-ownership world, we maintain that linearity that we started with as long as possible.  This requires plumbing through a 'linear state' value (a GADT probably) through *all* calls that use a buffer.  Constructing this state value has to be done by something that initialises the virtual hardware.

Note that this isn't a concrete interface proposal: I need to defer to @yallop's expertise in this matter as he has supervised a student project that worked in this area last year.

@hannesm:
> Performance optimization to reduce copying is something we can do in a later step (after evaluating the different proposals). My current intuition is that copying into the OCaml heap for received buffers is fine (and hardly measurable).

Unfortunately it is pretty measurable when you have lots and lots of buffers being sliced up and passed around.  The most obvious example is Cohttp _vs_ httpaf at the application level, but further compounded in the tcp/ip layers. 

Linearity is a really good approach if we can get it right -- it satisfies both performance and safety needs by taking advantage of a vital single-ownership invariant that is almost always a bug if violated dynamically.

gasche:
Linearity can be enforced at runtime by adding an indirection: instead of passing "the buffer" as an OCaml value, the values you share/copy are "a mutable pointer to the buffer". Whenever you decide to give up ownership of the buffer to someone else, they create their own mutable pointer and you rewrite your own pointer to NULL, so that any of your future accesses to the buffer will fail.

(This assumes that there is no OCaml function to just dereference those pointers and return the "raw pointer" as an OCaml structure; otherwise this can be leaked, the cat is out of the bag. But it's actually sort of easy to guarantee this with "buffer of bytes", as you can just provide abstract get/set/read/write/sub/blit functions.)

I am not familiar with the Mirage codebase, but my gut feeling is that untying your core data structures to "bigarrays as they are in the standard library" would help quite a bit with experimenting with other approaches and seeing what works. If it was my project (it isn't!), I would try to make the types of data blobs more abstract to make such a transition possible.

avsm:
@gasche wrote:
> I am not familiar with the Mirage codebase, but my gut feeling is that untying your core data structures from "bigarrays as they are in the standard library" would help quite a bit with experimenting with other approaches and seeing what works.

This is right, but the devil's in the performance detail :-) The introduction of the bigarray primitives in #5771 means that it's hard to move to a separate library that wraps allocated C buffers but that can still efficiently unmarshal values to OCaml. We either have to use a string, or ensure that the offsets within the C malloced memory are the same as a bigarray representation.

If there's an effective workaround for that, or perhaps a fix in the compiler to add support for unmarshalling from arbitrary memory regions that were externally allocated, then I would be heavily in favour of experimenting with new memory representations for Mirage IO.  I don't think that either `string` or `bigarray` are satisfactory for dealing with heavy IO, since they were both designed for other purposes.

dinosaure:
About experimentation, I did a big work between `bytes`/`string`/`bigarray` with _GADT_, _functor_ and abstract type here: https://github.com/dinosaure/buffet/tree/master/lib

From some discussions with @chambart, the best trade-off (performance/readability) to be able to be abstracted over a _buffer_ is the _functor_ way. But I don't have a benchmark on a large application (such as `ocaml-git`) to say: this way is the best one.

With @Drup, it seems than about question of abstraction, the best is to provide a [common interface](https://github.com/dinosaure/buffet/blob/master/lib/s.mli) and let the end-user to use specific implementation according to this interface.

About linearity, this problem seems to be common (specially for `ocaml-git`). A runtime-check seems to be reasonable but I'm afraid about the failure case where it can appear from anywhere without any context. I mean, with `ocaml-git`, we have this kind of problem and the hard time is when we try to find where we share the buffer (and where we should not). Bug pattern of a shared and compromised buffer is not so difficult to be understandable.

It's why I would like a static analyze about that instead a runtime-check.

About the choice of `bigarray` instead `bytes` (or vice-versa), clearly I'm not sure if it's really interesting. Indeed, in some situations, to use `bytes` instead `bigarray` can be faster in certain contexts but it's not true for all cases. `digestif` provides `digest` functions for `bytes`, `string` and `bigarray` and, for specific applications, `bigarray` is better than `bytes`, specially when `bigarray` never move in the heap (so we are able to do some parallelization stuff on it easily). So the choice really depends on what we want and it's why I prefer an common API with _cast_ operations between each kind of buffers.

This is my 2 cts :+1: .

gasche:
> About linearity, this problem seems to be common (specially for ocaml-git). A runtime-check seems to be reasonable but I'm afraid about the failure case where it can appear from anywhere without any context. I mean, with ocaml-git, we have this kind of problem and the hard time is when we try to find where we share the buffer (and where we should not). Bug pattern of a shared and compromised buffer is not so difficult to be understandable.
>
> It's why I would like a static analyze about that instead a runtime-check.

If you use dynamic enforcement of unique ownership, I think it wouldn't be too hard to have a debug mode that helps you find where the issue come from. As an example example, you could have an opt-in mode where, when you transfor ownership to someone else, you could capture a stack trace and place that trace in the "there is no buffer anymore" structure (instead of just NULL). If a later access fails because the buffer is gone, at that point you can then display the saved trace that pinpoints where exactly the value was given to someone else.

@avsm: the primtives are a good point. Two vague ideas coming to mind:

- You could still use other structures for experiments, and compare with "bigarrays without the primitives" to get a fair performance comparison (everything would be appropriately slower).
- If you wanted an efficient out-of-heap buffer *as a library*, instead of performing a C call, you could write the access primitives as functions in assembly code that follows the OCaml calling convention and link to that. (This is not exactly as cheap as an inline read/write instruction, but it should be pretty good). (This is a hacky idea and I'm sure you would find people excited to experiment with that.)

talex5:
> write-only - not yet sure whether this is really useful - I fail to find a use-case -- checksum computations (which require to read sth) on IP/TCP/UDP get in the way of marking a buffer write-only

write-only would be useful if the firewall was going to be updated to use the new mirage-net-xen. Since 1.10, mirage-net-xen is giving out a cstruct that is under the control of the remote (malicious) domain, and can therefore change at any time. If the firewall reads out any data that it has just written and depends on it any way, we likely have a security vulnerability. Linear types don't help here because we don't control the other domain.

( For example, if the firewall decided to route a packet to a client app VM, wrote the packet to the shared ring, and then looked at the written packet to generate the NAT rules for the return path. The app VM could change the source or destination address and thus add a bogus rule to the NAT. A bit far-fetched I guess, but I'd rather not have to worry about such things. )

hannesm:
@talex5 good point. so the firewall `read` a packet from some shared memory, processes the data (which includes in the NAT case some rewriting, and checksum re-computation), and writes it to another region of shared memory. IIRC the current API in our network stack writes (e.g. TCP and IP data) to the buffer, and then computes the checksum using the very same buffer, which in your shown firewall case should not be done (but instead, some temporary scratch buffer for checksum computation should be used).

@gasche / @dinosaure / @avsm I'd really like to avoid functorising over Cstruct/buffer all of our libraries. the way forward sounds like finding a common interface of different implementations (as @dinosaure pointed to his s.mli) -- which does not expose the `type t` -- upstream the interface, fix users of Cstruct that rely on internals, and then we can play around with different implementations. WDYT?

talex5:
@hannesm yes, although in the case the checksum it doesn't really matter. It's a bit strange that the destination can cause us to calculate a checksum for something other than the packet we intended to send, but the only thing we do with the checksum is to send it to the same destination anyway.

emillon:
Regarding capabilities, I agree that it's something that's missing in the ecosystem. Having immutable binary data is super useful for things like crypto code.

I think it would be easier to integrate if this was a new library implemented on top of cstruct, though (rather than changing cstruct). That way we could incrementally adapt (some) cstruct users to that.

Regarding using strings vs bigarrays, yes using malloc is probably wasting many cycles in mirage at the moment, but brings the important property that data does not move. The arguments I've heard in favor of that include that it makes it possible to:
- erase secrets (not done in practice, probably difficult to do at scale, and only brings value when uninitialized data is exposed to app code)
- have page-aligned buffers (from what I understand, it was more important in earlier version of mirage than it is now)

dinosaure:
> I think it would be easier to integrate if this was a new library implemented on top of cstruct, though (rather than changing cstruct). That way we could incrementally adapt (some) cstruct users to that.

In the details, this PR does not change `cstruct` nor implementation. Implementation of `Cstruct_cap` is only a _weird_ include of `cstruct_core`/`cstruct` and I just put a `cstruct_cap.mli`.

About linearity, of course, we need to change `Cstruct.t` to put a private token as @gasche said.

emillon:
Put it differently, could this be a separate project, or does it require access to cstruct's private API?

dinosaure:
It's more about maintainability of this kind of package where, for any update of `cstruct`, we should update `cstruct-with-cap` too (where they should share the same implementation). Then, I need to investigate but I suspect a regression performance in this way (but may be I'm wrong in favor to cross-optimization of `*.cmx{a}`.

I prefer this way where we completely keep the compatibility with the current status of `cstruct` and provide a sub-module which integrate capabilities (and ensure than we keep exactly the same semantic of functions) - and by this way, avoid a _duplication_ of packages.

hannesm:
I agree with @dinosaure in the (tentative) migration plan:
- provide an API with capabilities in version X
- modify users to the new API
- drop old API in version Y

Now, to start the migration process, the first step is to provide a capability-API, rather sooner than later. Removal of some internals from the API to allow non-Bigarray-based implementations is orthogonal to that AFAICT.

Your proposal of a separate opam package adds more complexity: we'd need to add it as dependency to all users, and drop the cstruct package entirely once the migration is done (or should the other opam package be renamed to cstruct, and require all users to change again?).

I don't understand your "erasing secrets" argument, you're aware that `Cstruct.create_unsafe` is around (also `Bigarray.Array0.create`)? In my perspective, "erasing secrets" is still a long way to go in MirageOS (and likely requires OCaml runtime changes -- if you're interested, you may enjoy reading https://blogs.akamai.com/2014/04/heartbleed-update-v3.html - which makes clear that you need to protect not only the raw key material, but also lots of derived material). FWIW, cstruct does not deal with page alignment.

emillon:
Oh, erasing secrets is definitely not my argument - only something that I've heard several times and that was missing here for completeness. I agree that it's not the important part to fix right now.

> FWIW, cstruct does not deal with page alignment.

But `io-page` relies on cstruct to do that at the moment, right? I don't know how important that library is, but it wouldn't work with strings.

dinosaure:
PR rebased on master, if we can start to talk about details (mostly on `cstruct_cap.mli`) 👍 !

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
+type 'a rd = < rd: unit; .. > as 'a
+type 'a wr = < wr: unit; .. > as 'a
+type 'a aligned = < aligned: unit; .. > as 'a
```
hannesm: This is not used anywhere, what is its purpose? I'd drop it in a first revision, and add alignment at a later point

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+type rdwr =  < rd: unit; wr: unit; >
+type ro = < rd: unit; >
+type wo = < wr: unit; >
+
+type uint8 = int
+type uint16 = int
+type uint32 = int32
+type uint64 = int64
+
+val create : int -> rdwr t
```
hannesm: does `create` zero out the memory? do we need `create_unsafe` as well (I'm ok with no `create_unsafe`)

hannesm: the functions below (`of_string` / `of_bytes`) have an `?allocator` argument -- either drop it there, or add that to `create` as well!?

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+val check_alignment : 'a rd t -> int -> bool
+
+val get_char : 'a rd t -> int -> char
+val get_uint8 : 'a rd t -> int -> uint8
+val set_char : 'a wr t -> int -> char -> unit
+val set_uint8 : 'a wr t -> int -> uint8 -> unit
+
+val sub : 'a rd t -> int -> int -> 'a rd t
+val shift : 'a rd t -> int -> 'a rd t
+val copy : 'a rd t -> int -> int -> string
```
hannesm: I'd suggest to name this `to_string`.

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+val blit : 'a rd t -> int -> 'b wr t -> int -> int -> unit
+val blit_from_string : string -> int -> 'a wr t -> int -> int -> unit
+val blit_from_bytes : bytes -> int -> 'a wr t -> int -> int -> unit
+val blit_to_bytes : 'a rd t -> int -> bytes -> int -> int -> unit
+
+val memset : 'a wr t -> int -> unit
+
+val len : 'a t -> int
+val set_len : 'a wr t -> int -> 'a wr t
+val add_len : 'a wr t -> int -> 'a wr t
```
hannesm: wow, I've not seen this before, and am not sure anyone is using it (GitHub search showed test cases)

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+val copy : 'a rd t -> int -> int -> string
+
+val blit : 'a rd t -> int -> 'b wr t -> int -> int -> unit
+val blit_from_string : string -> int -> 'a wr t -> int -> int -> unit
+val blit_from_bytes : bytes -> int -> 'a wr t -> int -> int -> unit
+val blit_to_bytes : 'a rd t -> int -> bytes -> int -> int -> unit
+
+val memset : 'a wr t -> int -> unit
+
+val len : 'a t -> int
```
hannesm: no `'a rd t` here?

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+
+val split : ?start:int -> 'a t -> int -> 'a t * 'a t
+
+val to_string : 'a rd t -> string
+val to_bytes : 'a rd t -> bytes
+
+val hexdump : 'a rd t -> unit
+val hexdump_to_buffer : Buffer.t -> 'a rd t -> unit
+val hexdump_pp : Format.formatter -> 'a rd t -> unit
+val debug : 'a rd t -> string
```
hannesm: I'd keep a `pp` function, and remove the other 3 above.

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,88 @@
...
+val copyv : 'a rd t list -> string
+val fillv : src:'a rd t list -> dst:'b wr t -> int * 'a rd t list
+
+type 'a iter = unit -> 'a option
+
+val iter : ('a rd t -> int option) -> ('a rd t -> 'v) -> 'a rd t -> 'v iter
+val fold : ('acc -> 'x -> 'acc) -> 'x iter -> 'acc -> 'acc
+val append : 'a rd t -> 'b rd t -> rdwr t
+val concat : 'a rd t list -> rdwr t
+val rev : 'a rd t -> rdwr t
```
hannesm: not entirely sure, but my intuition for `append` etc. would be to retain the capability, and not widen it.

talex5: Append always makes a copy, so widening it is the right thing to do (we should probably be clearer about this in the doc-string, though).

lib/cstruct_cap.mli
talex5: Yeah, I think we should remove these. Searching my local disk, I found that mirage-block-xen uses `add_len`:

https://github.com/mirage/mirage-block-xen/blob/master/lib/block_request.ml#L61

It would probably be better to move this operation into cstruct itself, e.g.

```ocaml
val coalesce : t -> t -> t option
(** [coalesce a b] is a cstruct that covers [a] and [b] if they belong to the same
    underlying array and [b] follows [a] directly in memory. *)
```

dinosaure: Waiting to merge #245

lib/cstruct_cap.mli
dinosaure: Agree with @talex5

dinosaure:
PR rebased, waiting #245 and about `?allocator` argument, I would like more feedback about that. Then, I will make documentation on `cstruct_cap.mli` in an other commit.

hannesm:
about `?allocator`: I've looked through my cloned repositories, and there is not a single use thereof -- i.e. for simplicity, I'd remove it from the new interface. If there's demand at a later point, we can always re-introduce such an argument. It's also not clear to me what its semantics should be, given that `t` is now abstract.

dinosaure:
All reviews was done, so we can start to write documentation if everyone agree with this interface 👍 !

avsm:
I think that it would be good to separate the allocating from the non-allocating portions of the interface as cleanly as possible.

Right now, using a few (rev, concat) allocate a fresh cstruct, which will in turn call the system malloc (a very expensive operation in comparison to an OCaml allocation).  My feeling is that cstruct should be only be used for inplace things, and for smaller values we should actually be copying them into an OCaml value instead and then eventually writing them back to a cstruct when they need to be transmitted onwards.

So in the new interface, I'd suggest keeping the ?allocator argument on the of_bytes/of_string since its very cheap there, but to do "something" about Cstruct.rev/concat forcing the consumer to allocate a new cstruct.  Perhaps a Cstruct.rev_to_buffer or something?  Need to look at the existing usecases of these functions to determine how they're used to see if this is a good idea.  Another one is to provide an output Cstruct for the results to go into, so that the caller can just allocate a single bigger cstruct for the results of the concat.

hannesm:
> Right now, using a few (rev, concat) allocate a fresh cstruct, which will in turn call the system malloc (a very expensive operation in comparison to an OCaml allocation). My feeling is that cstruct should be only be used for inplace things, and for smaller values we should actually be copying them into an OCaml value instead and then eventually writing them back to a cstruct when they need to be transmitted onwards.

ok, right now I find the interface inconsistent -- `of_bytes/of_string` are the only two values which accept an optional `allocator` argument since 6 years, though I still have not found any user. other functions which allocate, `concat/append/rev/of_hex` don't accept an allocator.

> So in the new interface, I'd suggest keeping the ?allocator argument on the of_bytes/of_string since its very cheap there, but to do "something" about Cstruct.rev/concat forcing the consumer to allocate a new cstruct. Perhaps a Cstruct.rev_to_buffer or something? Need to look at the existing usecases of these functions to determine how they're used to see if this is a good idea. Another one is to provide an output Cstruct for the results to go into, so that the caller can just allocate a single bigger cstruct for the results of the concat.

In the capabilities interface, the `type t` is private (IMHO it should be abstract, we may need to expose `to_bigarray` @dinosaure) - I don't quite get how a user would supply a `allocator : int -> t` in these circumstances. If you have compelling use cases at hand, please show them. You mentioned "output Cstruct for the result to go into" -- you can of course write such code:

```OCaml
let data = Cstruct_cap.create 100 in
Cstruct_cap.blit src 0 data 0 50 ;
Cstruct_cap.blit src2 0 data 50 50 ;
```

In respect to "cstruct should be only be used for inplace things, and for smaller values we should actually be copying them into an OCaml value": I've an idea to come up with a cstruct_cap implementation which actually uses string/bytes instead of bigarray. of course at the device layer, some may need to *copy* the data into page-aligned buffers. I'm interested in observing the performance behaviour between these two approaches.

avsm:
@hannesm:
> ok, right now I find the interface inconsistent -- of_bytes/of_string are the only two values which accept an optional allocator argument since 6 years, though I still have not found any user. other functions which allocate, concat/append/rev/of_hex don't accept an allocator.

That's right -- I used the allocator argument to do some tracing to reduce latency (for Jitsu iirc), and then the later functions were added subsequently.

But it is indeed tricky to make this sort of tracing work when the value is abstract. I just wanted to point out the importance of not locking us into having lots of small `malloc`s (which speeding up requires ever more ingenuity in a C layer). I'm suggesting that `Cstruct.concat` should for example be considered harmful, and should permit code like yours instead:

``` 
let data = Cstruct_cap.create 100 in
Cstruct_cap.blit src 0 data 0 50 ;
Cstruct_cap.blit src2 0 data 50 50 ;
```

Rewriting Cstruct.rev/concat in terms of combinators over Cstruct_cap seems quite powerful, particularly as we can then use your string/bytes alternatives instead of bigarrays.

dinosaure:
> Rewriting Cstruct.rev/concat in terms of combinators over Cstruct_cap seems quite powerful, particularly as we can then use your string/bytes alternatives instead of bigarrays.

What you have in your mind (in terms of interface)?

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,85 @@
...
+
+val check_bounds : 'a rd t -> int -> bool
+val check_alignment : 'a rd t -> int -> bool
+
+val get_char : 'a rd t -> int -> char
+val get_uint8 : 'a rd t -> int -> uint8
+val set_char : 'a wr t -> int -> char -> unit
+val set_uint8 : 'a wr t -> int -> uint8 -> unit
+
+val sub : 'a rd t -> int -> int -> 'a rd t
```
hannesm: the two `int` arguments should be labeled here as well!

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,85 @@
...
+val shift : 'a rd t -> int -> 'a rd t
+val to_string : ?off:int -> ?len:int -> 'a rd t -> string
+val to_bytes : ?off:int -> ?len:int -> 'a rd t -> bytes
+
+val blit : 'a rd t -> int -> 'b wr t -> int -> int -> unit
+val blit_from_string : string -> int -> 'a wr t -> int -> int -> unit
+val blit_from_bytes : bytes -> int -> 'a wr t -> int -> int -> unit
+val blit_to_bytes : 'a rd t -> int -> bytes -> int -> int -> unit
```
hannesm: labels would be nice in the above block of `blit` functions -- `src` `src_off` `dst` `dst_off` `length`

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,85 @@
...
+
+val memset : 'a wr t -> int -> unit
+
+val len : 'a rd t -> int
```
hannesm: I'd rename to `length`, as it is named all over the OCaml stdlib.

hannesm:
thanks again. IMHO this is good to get merged (but needs a rebase of cstruct.ml according to github), you mentioned that you intend to write some API docs as well. I'd really like to get this into a release of cstruct soon and start using it :)

dinosaure:
Rebased 👍

dinosaure:
Documentation was done. I don't want to push the merge button but it's ready to merge I believe.

lib/cstruct_cap.ml
```diff
@@ -0,0 +1,68 @@
...
+
+external ro : 'a rd t -> ro t = "%identity"
+external wo : 'a wr t -> wo t = "%identity"
+
+let of_string ?off ?len x =
+  Cstruct_core.of_string ?off ?len x
+let of_bytes ?off ?len x =
+  Cstruct_core.of_bytes ?off ?len x
+
+(* XXX(dinosaure): discard [?allocator] arguments (see discusion in #237). *)
```
hannesm: that's done :) you can remove that comment!

hannesm:
thanks for the documentation. imho this is good now API-wise.  it would be great to have some test cases (likely you can copy and modify some Cstruct tests for Cstruct_cap)!?

lib/cstruct_cap.ml
```diff
@@ -0,0 +1,68 @@
+include (Cstruct_core : module type of Cstruct_core with type t := Cstruct_core.t)
```
samoht: missing file headers

lib/cstruct_cap.mli
```diff
@@ -1,3 +1,19 @@
+(*
+ * Copyright (c) 2012-2014 Anil Madhavapeddy <anil@recoil.org>
```
hannesm: you should actually mention yourself and 2019 in here :)

dinosaure:
Who wants to push merge button :smiling_imp: ?

avsm:
I've got some fixes to the ocamldoc to push and then i'll merge

avsm:
Hm, the wrapping makes the Cstruct.Cap odoc look absolutely terrible :-/

I'm tempted to unwrap this but make Cstruct_cap and Cstruct_core into private modules so their cmis arent exposed.

avsm:
Good to merge after CI I think. This has now three toplevel modules: Cstruct, Cstruct_cap and Cstruct_core. Those are unlikely to module clash, and the docs are neater and there is no dependency between Cstruct and Cstruct_cap.

lib/cstruct_cap.mli
```diff
@@ -0,0 +1,376 @@
...
+(** Type of write capability. *)
+
+type 'a t
+(** Type of cstruct with capabilities ['a]. *)
+
+type buffer = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
+(** Type of buffer. A {!t} is composed of an underlying buffer. *)
+
+type rdwr =  < rd: unit; wr: unit; >
+(** Type of read-and-capability. *)
```
darioteixeira, line 31: Apologies for butting in. Shouldn't this be "read-and-write capability"? ;-)

PR #239, opened 2019-03-19
Sexplib is now an optional library for the base `Cstruct` module

### LINKED ISSUE #222: sexplib dependency
can we make the sexplib dependency optional (similar to how it was done in uri and ipaddr)!? thanks //cc @avsm (this is first for the main cstruct package, but same would be nice for the ppx_cstruct package - see #116).

avsm:
A new `Cstruct_sexp` module has been introduced with the serialiser functions, contained within the `cstruct-sexp` opam package.

To convert old code, simply use `Cstruct_sexp.t` instead of `Cstruct.t` in a record type for which you are using `[@@deriving sexp]`. This is a type alias to `Cstruct.t` but also has the right sexp-conversion functions in scope.  There is an example of this in the `ppx_test/with-sexp` directory in the source repo.

fixes #222, cc @TheLortex

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,25 @@
+v4.0.0
+------
+
+- Sexplib is now an optional library for the base `Cstruct` module.
+  A new `Cstruct_sexp` module has been introduced with the serialiser
+  functions, contained within the `cstruct-sexp` opam package.
+
+  To convert old code, simply use `Cstruct_sexp.t` instead of
+  `Cstruct.t` in a record type for which you are using `[@@deriving sexp]`.
+  This is a type alias to `Cstruct.t` but also has the right
+  sexp-conversion functions in scope.  There is an example of this
+  in the `ppx_test/with-sexp` directory in the source repo.
+
+  When you have converted and released your library, add an
+  opam constraint of `cstruct {>="4.0.0"}` to your own opam
+  packages to ensure that they pick up this version of the library.
+  (fixes #222, @avsm)
+
+- Use computed versions in opam files to ensure that dependent
+  opam packages such as cstruct-async get the same base version
+  of cstruct to avoid mismatches.
+
 v3.7.0 2019-03-10
 -----------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+v3.7.0 2019-03-10
+-----------------
+
+- Improve performance by not doing redundant bounds checks
+  in both the Bigarray and Cstruct level (#236 @Reperator @chambart)
+- Ignore fields starting wih `_` by skipping code generation
+  but still respecting the space usage of that field. This was
+  a convention before but is now enforced by the code generator
+  to save space in the output (#233 @emillon)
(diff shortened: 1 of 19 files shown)


dinosaure:
Happy to see this PR!

cstruct-sexp.opam
```diff
@@ -19,7 +19,7 @@ depends: [
   "ocaml" {>= "4.03.0"}
   "dune" {build & >= "1.0"}
   "sexplib" {< "v0.12"}
-  "cstruct" {=version}
+  "cstruct" {>= "3.6.0"}
```
dinosaure: Maybe too high ... But fix CI about `ppx_cstruct`. Without this constraint, OPAM wants to install `cstruct.0.4`.

cstruct-sexp.opam
```diff
@@ -0,0 +1,31 @@
...
+build: [
+  ["dune" "subst"] {pinned}
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
+]
+depends: [
+  "ocaml" {>= "4.03.0"}
+  "dune" {build & >= "1.0"}
+  "sexplib" {< "v0.12"}
+  "cstruct" {>= "3.6.0"}
```
dinosaure: Maybe too high ... But fix CI about `ppx_cstruct`. Without this constraint, OPAM wants to install `cstruct.0.4`.

avsm: yes thx, fixed in f9e1463f9df71d088f6e3dbb1d66202678a9fef5

dinosaure:
A `REVDEP` will break the world I think where several packages depends on `cstruct` but I approved this merge. /cc @emillon for review

avsm:
I'll constrain the revdeps in opam-repo in prep for this. I dont think a huge number of things use the sexp functionality

cstruct-sexp.opam
```diff
@@ -0,0 +1,31 @@
...
+  "alcotest" {with-test}
+]
+synopsis: "Sexp serialisers for C-like structures"
```
emillon: How about this?

```suggestion
synopsis: "S-expression serialisers for C-like structures"
```

CHANGES.md
```diff
@@ -1,3 +1,28 @@
+v4.0.0
+------
+
+- Sexplib is now an optional library for the base `Cstruct` module.
+  A new `Cstruct_sexp` module has been introduced with the serialiser
+  functions, contained within the `cstruct-sexp` opam package.
```
emillon: The ppx will emit code that depends on sexplib when using `[@@sexp]`. I think that this should be documented somewhere. A potential source of confution is that it sets `(ppx_runtime_libraries cstruct)` only - this is probably what we want, but it will miss the `sexplib` dependency if for some reason the generated code is the only use of sexplib in a module.

avsm: Yes I noticed this in the porting of the tests. The workaround was pretty easy (just depend on sexplib), but I'll note this explicitly in the README (which I have to update, just remembered I only did CHANGES so far)

Issue #235, opened 2019-03-07
ppx: duplicate fields are not detected


emillon:
Hi,

The following piece of code is not detected as an error:

```
[%%cstruct
type dup = {
  x : uint8_t;
  x : uint8_t;
} [@@little_endian]
]
```

(both accessors with the same name are emitted).

This can happen when copy/pasting, so we might want to catch that.

Thanks!

avsm:
Agreed, we should reject this.  If you have time to do a patch for the 4.0 bump (in #239) then this could also catch any libraries that do this in the revdeps checks for that.

emillon:
I should be able to do that this afternoon hopefully.

PR #240, opened 2019-03-20
Detect duplicate fields in %%cstruct types

emillon:
See #235

This can happen when copy/pasting, so it is helpful to detect that.

This includes a refactor so that ignored fields are first class, and also improves the error messages.

### DIFF
--- a/ppx_test/errors/cstruct_duplicate_field.ml.expected
+++ b/ppx_test/errors/cstruct_duplicate_field.ml.expected
@@ -1,2 +1,2 @@
-File "cstruct_duplicate_field.ml", line 2, characters 0-77:
+File "cstruct_duplicate_field.ml", line 5, characters 2-14:
 Error: ppx_cstruct error: field x is present several times in this type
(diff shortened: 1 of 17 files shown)


ppx_test/errors/cstruct_duplicate_field.ml.expected
```diff
@@ -0,0 +1,2 @@
+File "cstruct_duplicate_field.ml", line 2, characters 0-77:
+Error: ppx_cstruct error: field x is present several times in this type
```
avsm: the duplication of `error` here is a minor concern ;-) just drop the second `error`?

emillon: it's caused by `Location.raise_errorf ~loc ("ppx_cstruct error: " ^^ fmt)` so it's global, but I agree that it should be fixed

emillon:
@avsm this should be all good now if you want to have a look :+1:

emillon:
Yes... these kind of checks would be easier if `[%%cstruct]` was handled as a deriving plugin (and this would be caught by the compiler), but we can fix that in cstruct 5.x maybe :)

avsm:
yeah the reason it isnt a deriving plugin is that there's no actual type definition with a `cstruct`, only with `cenum`.

Issue #203, opened 2018-07-18
Add cstruct.js as a runtime file


copy:
Dune supports a mechanism to automatically find and link the JavaScript dependencies of libraries, so one doesn't need to specify `+cstruct/cstruct.js` manually.

See `(javascript_files (<files-list>))`: https://dune.readthedocs.io/en/latest/dune-files.html#js-of-ocaml

avsm:
This one's in cstruct 4.0.0

PR #241, opened 2019-03-20
Javascript packaging fixes

jonludlam:


jonludlam:
I'm not totally sure of the second patch - there are indications upstream that the javascript stubs are moving, but I'm not sure of the reasoning - https://github.com/ocsigen/js_of_ocaml/commit/7c1c8cff90712f6df67026ba7349941ae7ebf603#diff-6f8d707ac8ba0276a5f94701eb821566

avsm:
I was *just* wondering what the random `js/` directory that @djs55 added a few years ago was for, and here comes this patch :-)

avsm:
is there any way to have a testcase for this that could do something without a browser around?

jonludlam:
Yes - as long as we've got node around.

avsm:
sure -- could not hook this into the full test case and just run it manually before a release

avsm:
(and in CI)

emillon:
This will take care of #203. It looks like we could test this a bit by making a tiny js_of_ocaml project? (that wouldn't execute the code, but at least this can make sure that we can link everything properly)

emillon:
(but if the function names are not checked by `js_of_ocaml`, maybe it's not useful)

avsm:
The fix is right -- can do the testcase as a separate PR

PR #244, opened 2019-03-25
Update fuzz tests to released crowbar API

talex5:
In https://github.com/mirage/ocaml-cstruct/pull/237#discussion_r268570118, @hannesm noted that some cstruct functions can be used to access memory outside of the original cstruct (but inside the original buffer), which seems a bit surprising. I noticed that the fuzz tests actually check for the safer behaviour, so I tried running them to see why they were passing. It turns out that they have bit-rotted and no longer compile.

This PR updates the tests to use the released crowbar API. I ran afl-fuzz on fuzz.exe, after a few minutes it reported three failures:

```
sub 0 204
sub: FAIL

When given the input:

    [#1 [44943; 0; 0]; 0; 204]
    
the test failed:

    check false
```

That is:

```ocaml
utop # let whole = Cstruct.create 44943 in
       let empty = Cstruct.sub whole 0 0 in
       Cstruct.sub empty 0 204;;
- : t = {buffer = <abstr>; off = 0; len = 204}
```

```
add_len: FAIL

When given the input:

    [#1 [60159; 8; 0]; 32768]
    
the test failed:

    check false
```

```
set_len: FAIL

When given the input:

    [#1 [61056; 0; 0]; 512]
    
the test failed:

    check false
```

I'm not sure what we expect the correct behaviour to be here, but either the tests or the code should be changed.

### DIFF
--- a/fuzz/dune
+++ b/fuzz/dune
@@ -1,3 +1,3 @@
 (executable
  (name fuzz)
- (libraries cstruct crowbar fmt))
+ (libraries cstruct cstruct-sexp crowbar fmt))
--- a/fuzz/fuzz.ml
+++ b/fuzz/fuzz.ml
@@ -11,31 +11,33 @@ let create_sub x start len =
     Cstruct.sub c start len
   with Invalid_argument _ -> bad_test ()
 
-let cstruct = Choose [
-    Map ([int8], create);
-    Map ([range 0x10000; int; int], create_sub);
+let cstruct = choose [
+    map [int8] create;
+    map [range 0x10000; int; int] create_sub;
   ]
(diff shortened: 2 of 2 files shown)


hannesm:
NB: in https://github.com/mirage/mirage-tcpip/pull/394 i removed various `set_len` calls. IMHO `set_len` and `add_len` should be removed, and `sub` be fixed. thanks!

dinosaure:
Same as @hannesm, `sub` should be fixed. However, we need to see if `mirage-tcpip` relies on this behavior.

avsm:
thanks!

PR #245, opened 2019-03-25
Fixes for #244

hannesm:
see #244 -- `sub` does one additional check, and `set_len/add_len` are marked as deprecated. //cc @talex5

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -125,12 +125,46 @@ let debug t =
   ) else
     Format.asprintf "%a" pp_t t
 
-let sub t off0 len =
-  let off = t.off + off0 in
-  let new_end = off + len in
+let sub t off len =
+  (* from https://github.com/mirage/ocaml-cstruct/pull/245
+
+     Cstruct.sub should select what a programmer intuitively expects a
+     sub-cstruct to be. I imagine holding out my hands, with the left
+     representing the start offset and the right the end. I think of a
+     sub-cstruct as any span within this range. If I move my left hand only to
+     the right (new_start >= t.off), and my right hand only to the left
+     (new_end <= old_end), and they don't cross (new_start <= new_end), then I
+     feel sure the result will be a valid sub-cstruct. And if I violate any one
+     of these constraints (e.g. moving my left hand further left), then I feel
+     sure that the result wouldn't be something I'd consider to be a sub-cstruct.
+
+     Wrapping considerations in modular arithmetic:
+
+     Note that if x is non-negative, and x + y wraps, then x + y must be
+     negative. This is easy to see with modular arithmetic because if y is
+     negative then the two arguments will cancel to some degree the result
+     cannot be further from zero than one of the arguments. If y is positive
+     then x + y can wrap, but even max_int + max_int doesn't wrap all the way to
+     zero.
+
+     The three possibly-wrapping operations are:
+
+     new_start = t.off + off. t.off is non-negative so if this wraps then
+     new_start will be negative and will fail the new_start >= t.off test.
+
+     new_end = new_start + len. The above test ensures that new_start is
+     non-negative in any successful return. So if this wraps then new_end will
+     be negative and will fail the new_start <= new_end test.
+
+     old_end = t.off + t.len. This uses only the existing trusted values. It
+     could only wrap if the underlying bigarray had a negative length!  *)
+  let new_start = t.off + off in
+  let new_end = new_start + len in
   let old_end = t.off + t.len in
-  if new_end > old_end || off < off0 || new_end < off || off0 < 0 || len < 0 || not (check_bounds t new_end) then err_sub t off0 len
-  else { t with off; len }
+  if new_start >= t.off && new_end <= old_end && new_start <= new_end then
+    { t with off = new_start ; len }
+  else
+    err_sub t off len
 
 let shift t amount =
   let off = t.off + amount in
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -127,9 +127,9 @@ let debug t =
 
 let sub t off0 len =
   let off = t.off + off0 in
-  let offlen = off + len in
+  let new_end = off + len in
   let old_end = t.off + t.len in
-  if offlen > old_end || off < off0 || offlen < off || off0 < 0 || len < 0 || not (check_bounds t (offlen)) then err_sub t off0 len
+  if new_end > old_end || off < off0 || new_end < off || off0 < 0 || len < 0 || not (check_bounds t new_end) then err_sub t off0 len
   else { t with off; len }
 
 let shift t amount =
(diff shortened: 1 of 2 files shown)


lib/cstruct.ml
```diff
@@ -128,7 +128,8 @@ let debug t =
 let sub t off0 len =
   let off = t.off + off0 in
   let offlen = off + len in
-  if off < off0 || offlen < off || off0 < 0 || len < 0 || not (check_bounds t (offlen)) then err_sub t off0 len
+  let old_end = t.off + t.len in
+  if offlen > old_end || off < off0 || offlen < off || off0 < 0 || len < 0 || not (check_bounds t (offlen)) then err_sub t off0 len
```
talex5: Can we simplify this check? If it's a sub-cstruct of the input, it must be within the bigarray too.

Perhaps something like:

```ocaml
let sub t off0 len =
  let new_start = t.off + off0 in
  let new_end = new_start + len in
  let old_end = t.off + t.len in
  if new_start <= new_end && new_start >= t.off && new_end <= old_end then { t with off = new_start; len }
  else err_sub t off0 len
```

lib/cstruct.mli
```diff
@@ -335,12 +335,16 @@ val len: t -> int
 val set_len : t -> int -> t
 (** [set_len t len] sets the length of the cstruct [t] to a new absolute
     value, and returns a fresh cstruct with these settings.
-    @raise Invalid_argument if [len] exceeds the size of the buffer. *)
+    @raise Invalid_argument if [len] exceeds the size of the buffer.
+
+    @deprecated This function is deprecated. *)
```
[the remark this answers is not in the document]
hannesm: oh I used the other `@deprecated` occurences in the same file as example, good point to use a proper annotation!

lib/cstruct.ml
hannesm: maybe, I'd leave this as a potential future PR.

avsm:
This looks correct to me, but I'd like to get it into cstruct.4.1.0 after all the sexp build changes have gone out, since mirage-tcpip needs testing with this changeset as well.

talex5:
Other uses of `set_len` and `add_len` (from https://github.com/search?q=org%3Amirage+add_len&type=Code):

- https://github.com/mirage/mirage-block-xen/blob/master/lib/block_request.ml#L67 ("merge adjacent cstruct buffers"; add new cstruct function for this?)
- https://github.com/mirage/ocaml-git/blob/master/src/git/helper.ml#L219 (don't know what this does - @dinosaure?)
- https://github.com/mirage/ocaml-vmnet/blob/38d156580f55404d25d5b76e3343aa418c5ccf56/lib/vmnet.ml (possibly using it as a `truncate`? maybe add a non-deprecated `truncate` function for that?)
- https://github.com/mirage/ocaml-openflow/blob/dcda113745e8edc61b5508eb8ac2d1e864e1a2df/lib/ofsocket.ml (dead? no updates since 2013)
- https://github.com/mirage/mirage-tcpip/blob/ee3e8a3335c620e311a6d36d55539294a2c64ca6/src/tcp/tcp_packet.ml (used as truncate)

lib/cstruct.ml
talex5: I think it should have a TODO at least, to avoid confusing future readers.

hannesm:
@avsm imho the cstruct.sub check is critical, and should be in the next release. I'm also very much in favour to get #237 in the next release.
I would rather wait some days for a release than to hurry one which has (a) breaking changes and (b) is still unsafe and (c) has no sufficiently convincing API for read-only/write-only cstructs.

lib/cstruct.ml
avsm: @hannesm the one above is easier to read -- why not just use that one instead, since we've got to test a bunch of uses of `sub` anyway with whichever one we go with

avsm:
@hannesm sorry, no -- I don't like coupling semantic changes with build-breaking changes if at all possible. I've cut cstruct.4 specifically for the sexp changes and for upper bounds in opam-repo, and that can be tested independently of the sub changes here.  We can cut a point release with the sub change for a cstruct.3.x release too, if you deem it critical enough.

hannesm:
@avsm as you like, good luck.

lib/cstruct.ml
talex5: It's good to be careful here. For speed, we don't do any further bounds checking after creating a cstruct, so we need to get this right. Here is my argument for the correctness of the function above.

`Cstruct.sub` should select what a programmer intuitively expects a sub-cstruct to be. I imagine holding out my hands, with the left representing the start offset and the right the end. I think of a sub-cstruct as any span within this range. If I move my left hand only to the right (`new_start >= t.off`), and my right hand only to the left `new_end <= old_end`, and they don't cross `new_start <= new_end`, then I feel sure the result will be a valid sub-cstruct. And if I violate any one of these constraints (e.g. moving my left hand further left), then I feel sure that the result wouldn't be something I'd consider to be a sub-cstruct.

Therefore, if this code returns a new cstruct then `new_start` and `new_end` will be in range, and I think it's clear enough that they will also be what the user expected.

However, we don't store `new_end`; we instead store `len`, from which `new_end` is calculated. Is it possible to provide an out-of-range `len` that still results in an in-range `new_end`, given that we are using modular arithmetic? No. For any valid `new_end` there must be some correct `len`, and `new_end = new_start + len` has only one solution for `len` given `new_end` and `new_start`.

(The manual says `All operations are taken modulo 2^Sys.int_size` so we don't have to worry about crazy C-style Undefined Behaviour). 

If the above isn't convincing, there is another way to see that modular arithmetic doesn't cause a problem: if the function returns a new cstruct then none of the three addition operations in the function can have wrapped.

Note that if `x` is non-negative, and `x + y` wraps, then `x + y` must be negative. This is easy to see with modular arithmetic because if `y` is negative then the two arguments will cancel to some degree the result cannot be further from zero than one of the arguments. If `y` is positive then `x + y` can wrap, but even `max_int + max_int` doesn't wrap all the way to zero.

The three possibly-wrapping operations are:

`new_start = t.off + off0`. `t.off` is non-negative so if this wraps then `new_start` will be negative and will fail the `new_start >= t.off` test.

`new_end = new_start + len`. The above test ensures that `new_start` is non-negative in any successful return. So if this wraps then `new_end` will be negative and will fail the `new_start <= new_end` test.

`old_end = t.off + t.len`. This uses only the existing trusted values. It could only wrap if the underlying bigarray had a negative length!

It would be good if someone could check the reasoning above and/or provide a simpler argument that the alternative code is correct.

dinosaure:
Hmmhmm about `ocaml-git` ... I think a dragoon is hidden behind. I will investigate soon this way to unserialize/serialize Git object, however it will be not easy to remove it at this time.

hannesm:
NB: I moved the add_len/set_len changes to #251.

Here, I force-pushed @talex5 code, and included his very nicely written text as a comment to sub. This works fine, and I follow Thomas' description (also the wrapping behaviour). Thanks for that. IMHO good to get merged.

I tested a locally pinned cstruct with this PR and started a `static_website_tls` unikernel. It works fine (serving websites via TLS, replying to ICMP, ...). This does not mean that there are any users of `Cstruct.sub` that try to widen the view -- we'll find them eventually ;)

PR #247, opened 2019-03-27
Use bigarray-compat

TheLortex:
The bigarray-compat package allows to selectively choose Bigarray or Stdlib.Bigarray according to ocaml version. Stdlib.Bigarray allows to not depend on unix for OCaml >= 4.07.

### DIFF
--- a/cstruct.opam
+++ b/cstruct.opam
@@ -18,6 +18,7 @@ build: [
 depends: [
   "ocaml" {>= "4.03.0"}
   "dune" {build & >= "1.0"}
+  "bigarray-compat"
   "alcotest" {with-test}
 ]
 synopsis: "Access C-like structures directly from OCaml"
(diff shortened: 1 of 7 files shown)


TheLortex:
This actually breaks a lot of revdeps that directly use `Bigarray` without explicitely mentioning it in their dune file. But this is important for mirage dunification.
@avsm what can we do ?

hannesm:
I'd propose to get this merged and released, and adjusting upper bounds of all the failed packages that use Bigarray directly (then, on a case-by-case basis find a solution for the specific package).

appveyor complained about "no package bigarray-compat" (likely it used an old opam-repo), a restart should succeed (but I can't find a restart button).

avsm:
I agree. I'll put a big note in the CHANGES file about this.

Issue #128, opened 2017-03-14
equal/compare for cenums?


yomimono:
It would be nice to automatically get `equal` or `compare` functions for types created from `cenum`s.

avsm:
is this for satisfying a particular function signature? (i.e. do you need just `compare` or is `equal` also definitely needed).  Since they are just integers, polymorphic comparison is normally fine, but generating explicit comparators is definitely a good thing.

yomimono:
In short, stuff potentially from cenums -> record -> alcotest .

avsm:
This sounds like having `compare` is enough -- I'm a little wary of generating too many functions from the ppx.

yomimono:
Sure, `compare` is probably fine.

emillon:
N.B: supposing that you want to compare by id, polymorphic comparison gives the wrong results for cases like this:

```ocaml
[%%cenum
type reversed = 
  | ONE_R [@id 2]
  | TWO_R [@id 1]
  [@@uint8_t]
] 
```

I'll have the ppx handle this, but that could be a source of bugs.

avsm:
good catch!

avsm:
fixed, will be in cstruct 5.0+

PR #248, opened 2019-03-29
ppx: generate compare functions for cenums

emillon:
Hi,

This is a patch for #128. This generates an extra `compare_X` function for each cenum. The comparison function corresponds to the id order.

Let me know what you think @yomimono @avsm.

Thanks!


yomimono:
I like it!

avsm:
lgtm!

PR #250, opened 2019-03-29
Extend the fuzz tests a little

talex5:
@yomimono : can we configure these to run automatically on CI somehow?

### DIFF
--- a/fuzz/fuzz.ml
+++ b/fuzz/fuzz.ml
@@ -7,7 +7,10 @@ let create x =
 
 let create_sub x start len =
   try
-    let c = Cstruct.create x in
+    let c = Cstruct.create_unsafe x in
+    for i = 0 to len - 1 do
+      Cstruct.set_uint8 c i i
+    done;
     Cstruct.sub c start len
   with Invalid_argument _ -> bad_test ()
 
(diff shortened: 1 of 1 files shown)


fuzz/fuzz.ml
```diff
@@ -51,48 +61,81 @@ let () =
...
+      match Cstruct.split ?start base len  with
+      | c1, c2 ->
+        check_within ~base c1;
+        check_within ~base c2;
+        let start = match start with None -> 0 | Some x -> x in
+        check (Cstruct.equal (Cstruct.sub base start len) c1);
+        check (Cstruct.equal (Cstruct.shift base (start + len)) c2)
       | exception Invalid_argument _ -> ()
     );
   add_test ~name:"BE.set_uint64" [cstruct; int] (fun c off ->
```
emillon: What do you think about adding similar cases for other BE setters and getters (where there are similar bound checks)?

yomimono:
Yes, but it's a bit tricky because (last I checked) the default images didn't include the +afl variants, so I had to make custom OCaml images with instrumentation enabled to get this working for other repositories and I no longer have access to that tooling.  It should be possible though; I'm looking into it now and will submit a PR once I have something working.

talex5:
The `fillv` test fails somewhat randomly for me, producing example cases that pass when run manually or are not even valid tests. I simplified it to this (which causes afl-fuzz to report crashes after a few seconds):

```ocaml
open Crowbar

let create _ = Bigarray.(Array1.create char c_layout) 0x10000

let cstruct = map [bool] create

let () =
  assert (Array.length Sys.argv = 2);   (* Prevent accidentally running in quickcheck mode *)
  add_test ~name:"fillv" [list cstruct] (fun src ->
      ignore src;
      (* Gc.full_major () *)
    )
```

Uncommenting the `Gc.full_major ()` fixes the problem, so I suspect this is some dodgy memory handling related to bigarrays not triggering GCs soon enough.

avsm:
All the variant compilers are in the `ocaml/opam2:<version>` for that particular compiler. e.g.:

```
$ docker run -it ocaml/opam2:4.07 opam switch
#   switch                      compiler                                     description
->  4.07                        ocaml-base-compiler.4.07.1                   4.07
    4.07+afl                    ocaml-variants.4.07.1+afl                    4.07+afl
    4.07+default-unsafe-string  ocaml-variants.4.07.1+default-unsafe-string  4.07+default-unsafe-string
    4.07+flambda                ocaml-variants.4.07.1+flambda                4.07+flambda
    4.07+force-safe-string      ocaml-variants.4.07.1+force-safe-string      4.07+force-safe-string
```

so you should be able to run it in CI.   It might be useful to run them in cloud.drone.io btw -- that has a very fast bare metal CI including arm64 (see https://github.com/ocaml/ocaml-beta-repository for an example) that might be interesting for cstruct fuzzing on non-x86 as well as x86

fuzz/fuzz.ml
talex5: I think anyone who has an idea for extra fuzz tests should feel free to make a PR adding them.

talex5:
I added a GC call to fix the failing test. @yomimono are you happy to merge this now, or do you want to set up the fuzzing CI first?

yomimono:
Let's merge this now, and I'll submit a separate PR for automated fuzz magic.

PR #256, opened 2019-04-16
Remove ppx_driver test dependency

emillon:
It shouldn't be necessary.


emillon:
All green @avsm - should I open a similar PR against opam-repository?

avsm:
Weird, wonder how it crept in.  A PR to opam-repo would be great -- just the latest version is fine since it did no harm previously but the constraint only affects 4.0+

PR #255, opened 2019-04-15
Add Windows/macOS/Linux and x86/arm64 fuzzing CI

avsm:
Azure Pipelines and cloud.drone.io

For the fuzz tests, this fixes #252, but how do we tell `afl-fuzz` how long to run for?  (See .drone.yml, tweaks welcome to the command line). It's running atm on https://cloud.drone.io/mirage/ocaml-cstruct/6/1/2 which I hope is a public URL.

PR #257, opened 2019-04-18
Suppress deprecation warnings in tests

emillon:
This ensures that `make test` works.

And these tests will be deleted with the corresponding functions are.

### DIFF
--- a/fuzz/fuzz.ml
+++ b/fuzz/fuzz.ml
@@ -115,7 +115,7 @@ let () =
       check (Cstruct.get_uint8 c i = x land 0xff)
     );
   add_test ~name:"set_len" [cstruct; int] (fun base len ->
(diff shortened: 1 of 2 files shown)


hannesm:
this looks fine to me, and is less global than c7b5a7e / 6498366

another option would be to get rid of the tests for the deprecated functions.... ;)

emillon:
Yes I'm pretty sure any global annotation is never going to be removed :)

I don't think the tests should be removed, there are still some uses of these functions and even if we planned to remove them, they might still change a bit so it's useful to keep the safety net.

hannesm:
ok, fine with me

avsm:
Oops, this clashes with #255 , will rebase this post that

Issue #263, opened 2019-07-10
ppx_cstruct testsuite broken under OCaml 4.08


gasche:
The `ppx_cstruct` testsuite is broken under OCaml 4.08.0. See

https://github.com/ocaml/opam-repository/pull/14464
https://ci.ocaml.org/log/saved/docker-run-568738f3749bf66f881cf36d58071286/a1d58c3c528c836550875fd71cad87479128b5a4
https://github.com/gasche/opam-repository/commit/1e216a1bb02ab0573571c70d78fe5b03c6f92056

Quoting the commit message from https://github.com/gasche/opam-repository/commit/1e216a1bb02ab0573571c70d78fe5b03c6f92056

> With 4.08 the compiler prints errors differently by default,
which appears to break expect-style tests that are checking for errors.
>
> This should probably be fixed at the upstream level of dune (probably
by calling the compiler with the '-error-style short' option in
tests), so in the future (if "fixed" versions of dune are released)
the (ocaml:version < 4.08) constraint on tests should be replaced by
(dune:version >= fixed-version) to re-enable testing.

It may be enough to enable `-error-style short` or `OCAML_ERROR_STYLE=short` from the `dune` fields of the ppx_cstruct test, but I wonder if a dune-level fix would be possible.

gasche:
(I'm not sure I have a good enough handle of the package-level vs. dune-level issue to report an issue on the Dune bugtracker right now. People that are more familiar with the mirage test setup, please feel free to give an opinion on what's the best fix and, if "upstream" is the answer, open an issue there yourself.)

emillon:
Thanks for the report. That looked fragile indeed. Adding first class support to this in dune looks reasonable, I'll bring that up.

avsm:
Sadly, it turns out that 4.09 has also further changed the error format, so simply setting CAML_ERROR_STYLE isn't fully sufficient. https://github.com/ocaml/opam-repository/pull/15381

(the short error style does work for 4.08 and below)

PR #265, opened 2019-09-02
len attribute should be an int literal

### LINKED ISSUE #264: silent failure on variable/expressions in [@len]
When there are fields with lengths (like `ephemeral: uint8_t [@len 32]` for example like in a cstruct), if you put a variable or an expression where the `32` is now, the ppx doesn't even give an error or complain at all, I think it just ignores the [@len ...]. 

This is pretty misleading and might be something to consider adding an error for? I just kept getting weird errors that the blit function I wanted didn't exist, and it took me a minute to trace it back to this. If someone was only using get and set this could lead to scary and difficult-to-detect memory bugs.

emillon:
Closes #264

This rejects things like `[@len ""]` and `[@len 1l]` which were previously silently ignored.

### DIFF
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -527,12 +527,17 @@ let constr_enum = function
     loc_err loc "invalid cenum variant"
 
 let get_len = function
-  | [ ({txt = "len"; _},
+  | [ ({txt = "len"; loc},
        PStr
          [{pstr_desc =
              Pstr_eval ({pexp_desc = Pexp_constant (Pconst_integer (sz, None)); _}, _)
-          ; _}])] ->
-    Some (int_of_string sz)
+          ; _}])]
+    ->
+    let n = int_of_string sz in
+    if n > 0 then
+      Some n
+    else
+      loc_err loc "[@len] argument should be > 0"
   | [{txt = "len"; loc}, _ ] ->
     loc_err loc "[@len] argument should be an integer"
   | _ ->
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -526,15 +526,20 @@ let constr_enum = function
   | {pcd_loc = loc; _} ->
     loc_err loc "invalid cenum variant"
 
+let get_len = function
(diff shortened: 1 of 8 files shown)


emillon:
ppxlib has some combinators to avoid this kind of boilerplate - might be worth investigating if we split the ppx out of the main cstruct package.

avsm:
I'd be delighted to get a ppx/ppxlib port of ppx_cstruct..

ppx/ppx_cstruct.ml
```diff
@@ -526,15 +526,20 @@ let constr_enum = function
   | {pcd_loc = loc; _} ->
     loc_err loc "invalid cenum variant"
 
+let get_len = function
+  | [ ({txt = "len"; _},
+       PStr
+         [{pstr_desc =
+             Pstr_eval ({pexp_desc = Pexp_constant (Pconst_integer (sz, None)); _}, _)
+          ; _}])] ->
+    Some (int_of_string sz)
```
avsm: It's probably worth checking that it's a positive, non-zero integer here.

ppx/ppx_cstruct.ml
```diff
@@ -526,15 +526,20 @@ let constr_enum = function
...
+  | [{txt = "len"; loc}, _ ] ->
+    loc_err loc "[@len] argument should be an integer"
```
avsm:
```suggestion
    loc_err loc "[@len] argument should be a positive, non-zero integer"
```

ppx/ppx_cstruct.ml
emillon: sure, done!

ppx/ppx_cstruct.ml
emillon: there are two check sites so I used this more specific version only when needed, how does that look?

avsm:
Looks good to me! Unrelated to this PR, but the tests fail on 4.08 since ppx_cstruct generates references to stdlib. I'll fix that in a separate PR.

emillon:
Yes that's #263 - the environment variable should do it. Thanks!

PR #266, opened 2019-09-22
PPX: Use Stdlib instead of Pervasives for generated enum compare functions

cypheon:
Currently, when using the `[%%cenum...]` PPX extension on OCaml 4.08, compilation fails with the error:

```
Error (alert deprecated): module Stdlib.Pervasives
Use Stdlib instead.

If you need to stay compatible with OCaml < 4.07, you can use the 
stdlib-shims library: https://github.com/ocaml/stdlib-shims
```

This change switches the generated code to use Stdlib.compare instead of Pervasives.compare. Of course, this will break compatibility in OCaml < 4.07, but as noted in the error message, users of older compilers could use the stdlib-shims.

Depending on your compatibility plans, a note should be added to the README. Maybe it's possible to add this directly to the ppx_cstruct dependencies in the .opam file (with a version restriction on the compiler)?


avsm:
Thanks for this. I'm tempted to just add `stdlib-shims` to the ppx-runtime dependencies of cstruct.

avsm:
fixed in #269, thanks!

PR #275, opened 2020-06-07
Upgrade ocaml-migrate-parsetree to support latest ocaml

bikallem:
This PR upgrades the ocaml-migrate-parsetree to the latest version supporting ocaml up to version 4.11.

The PR also upgrades dune to v2.0 to address some dune warnings. 

All tests pass.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,5 @@
-(lang dune 1.0)
+(lang dune 2.0)
+
+(allow_approximate_merlin)
+
 (name cstruct)
--- a/ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
+++ b/ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
@@ -10,5 +10,6 @@ type foo64 =
   [@@uint64_t]
 ]
 
-let foo = Lwt_main.run (Lwt.return ())
-
+let foo = 
+  let%lwt foo = Lwt.return () in 
+  Lwt.return foo
--- a/ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
+++ b/ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
@@ -10,5 +10,5 @@ type foo64 =
   [@@uint64_t]
 ]
 
-let%lwt foo = Lwt.return ()
+let foo = Lwt_main.run (Lwt.return ())
 
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,15 +6,15 @@ platform:
   arch: amd64
 
 steps:
-- name: build
-  image: ocaml/opam2:4.08
-  commands:
-  - sudo apt-get update && sudo apt-get -y install afl
-  - sudo chown -R opam .
-  - git -C /home/opam/opam-repository pull origin && opam update
-  - opam switch 4.08+afl
-  - opam pin add -n .
-  - opam depext -u cstruct-async cstruct-lwt cstruct-unix cstruct ppx_cstruct
-  - opam install -y .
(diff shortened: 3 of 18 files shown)


dune-project
```diff
@@ -1,2 +1,5 @@
-(lang dune 1.0)
+(lang dune 2.0)
```
avsm, line 1: This needs the ppx_cstruct.opam to have a corresponding `dune {>= "2.0.0"}` entry.

dune-project
```diff
@@ -1,2 +1,5 @@
...
+
+(allow_approximate_merlin)
```
avsm, line 3: This is odd; the ppx_cstruct should be the only thing in that directory. Can we get rid of it, or did you find you needed it to get the build to compile without warnings?

ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
```diff
@@ -10,5 +10,5 @@ type foo64 =
   [@@uint64_t]
 ]
 
-let%lwt foo = Lwt.return ()
+let foo = Lwt_main.run (Lwt.return ())
```
avsm, line 13: Why change this? That use of `let%lwt` was there in the test suite was there to check that ppx_cstruct and ppx_lwt work together.

dune-project
bikallem: yep, I get the following warning without that feature in ```dune-project```,
```
File "ppx_test/errors/dune", line 6, characters 3-60:
6 |    (run %{bin:cppo} -V OCAML:%{ocaml_version} %{input-file})))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning: .merlin generated is inaccurate. Cannot mix preprocessed and non
preprocessed specificiations.
Split the stanzas into different directories or silence this warning by
adding (allow_approximate_merlin) to your dune-project.
```

ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
bikallem: The recent version of ppx_lwt gives the warning below,
```
File "ppx_test/with-lwt/ppx_cstruct_and_lwt.ml", line 13, characters 0-27:
Error (warning 22): let%lwt should not be used at the module item level.
Replace let%lwt x = e by let x = Lwt_main.run (e)
```
I have now modified the test to use ```let%lwt``` inside rather than at the module toplevel to get around the warning.

dune-project
bikallem: Done.

dune-project
avsm: Can fix that in a future revision if necessary. Thanks for checking!

ppx_test/with-lwt/ppx_cstruct_and_lwt.ml
avsm: Good call; there should be no interop problems with modern ppx, but there were in the past, so its good to keep the test around.

dune-project
avsm: Sorry, just one last change needed here before merge. Since the `dune-project` affects all the associated projects, every opam package needs a corresponding `dune >2` entry. e.g. the Drone CI is failing as its trying to install `cstruct` with dune 1.11 and that no longer works.

bikallem: I have now added dune {>= "2.0.0"} constraint to all .opam files in repo.

bikallem:
@avsm all checks are green. I think it is ready to merge.

avsm:
Thanks! Releasing shortly.

PR #277, opened 2020-07-21
Add a middle-step to decode a format

dinosaure:
This PR wants to integrate what is available for [`astring`](https://github.com/dbuenzli/astring) but for `bigarray`/`cstruct`. I think, we missed a middle-step to decode a simple (or not really specified) format.

In the case where:
1) we are able to entirely load contents (with `mmap` for example)
2) we don't have a clear specification of the format (eg. ABNF)

It is interesting to take the advantage of the cheap underlying record of `Cstruct.t` to _parse_ the given contents. On my side, it appears for some cases:
- An example in the documentation is given about the Git tree object
  In `ocaml-git`, it appears that the most common case is to entirely load the Git object (apply an inflation) and decode it as is without the idea of a _kontinuation_ which will load the next chunk. The object is small enough to fit into a large `Cstruct.t` and the [`prompt`](https://github.com/inhabitedtype/angstrom/blob/master/lib/angstrom.ml#L139) step with `angstrom` is unnecessary (but still exists and costs). This API wants to provide (as `astring`) a way to _parse_ contents without an handler about input.

- A format such as the `file`'s database (which is not specified) 
  Some formats are under-specified and small iterations on it during the development is needed to finally be able to compute this kind of format. Tools such as `menhir` are less flexible is this context where we must solve _reduction/shift_ action. Something where the ambiguity is not a problem (or inherent from the format) is needed and I think this API is best for such case.

- Currently, some of my implementation still use `bytes`/`string` due to `astring` where `cstruct` (or `bigarray`) don't provide such API. It is the case about `colombe` or the new implementation of the Smart protocol into `ocaml-git`. I think it can help about an _unification_ of buffers into our eco-system (where the choice will be lead by technical points instead a lack of the eco-system).

So I would like to know the idea of this PR. I added documentation/tests/fuzzers and if the idea is cool for us, we can complete the API with some others functions if people want. (/cc @mirage/core).

PS: the code comes from the @dbuenzli's `astring` package of course with small modifications. Copyrights are respected.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Packages
       run: opam depext -yt cstruct
     - name: Dependencies
-      run: opam install cstruct --deps-only
+      run: opam install -t cstruct --deps-only
     - name: Build
       run: opam exec -- dune build -p cstruct
     - name: Test
--- a/fuzz/fuzz.ml
+++ b/fuzz/fuzz.ml
@@ -33,7 +33,7 @@ let check_within ~base x =
   check Cstruct.(x.len >= 0 && x.len <= base.len)
 
 let () =
-  assert (Array.length Sys.argv = 2);   (* Prevent accidentally running in quickcheck mode *)
+  (* assert (Array.length Sys.argv = 2);   (* Prevent accidentally running in quickcheck mode *) *)
   add_test ~name:"blit" [cstruct; int; cstruct; int; int] (fun src srcoff dst dstoff len ->
       try Cstruct.blit src srcoff dst dstoff len
       with Invalid_argument _ ->
@@ -115,7 +115,7 @@ let () =
       guard (i >= 0 && i < Cstruct.len c);
       Cstruct.memset c x;
       check (Cstruct.get_uint8 c i = x land 0xff)
-    );
+     );
   add_test ~name:"set_len" [cstruct; int] (fun base len ->
       match[@ocaml.warning "-3"] Cstruct.set_len base len with
       | x ->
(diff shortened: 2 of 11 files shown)


samoht:
Could you just add these function directly to `cstruct.ml`? Not sure we need an other module for these useful functions (and I really don't think we should add another sub-library)

dinosaure:
Yes, of course, I agree with you.

avsm:
Agreed about the same module. Also, if this is common to Astring and Cstruct now, could we have a module type somewhere for the common functions?

dinosaure:
The PR was updated to integrate all functions into the `Cstruct` module directly (and remove the sub-module).

dinosaure:
> Agreed about the same module. Also, if this is common to Astring and Cstruct now, could we have a module type somewhere for the common functions?

I can do a:
```ocaml
module type S = sig
  ...
end

include (S with type t := Cstruct.t)
```

If you think it's good (including the documentation). An other solution should be to add a /meta/ package which define such interface.

avsm:
It depends how useful such an interface might be. Do you think you'll ever use it to abstract across bigarray/string?  If not, it's not worth the trouble to pull out -- consumers can use 'module type of' if they need to.

hannesm:
hmm, is there any plan for `Cstruct_cap` and these helpers? or is `Cstruct_cap` an interface to be discarded even before being widely used?

dinosaure:
> It depends how useful such an interface might be. Do you think you'll ever use it to abstract across bigarray/string? If not, it's not worth the trouble to pull out -- consumers can use 'module type of' if they need to.

I think it can be interesting to switch and see performance impact between `string`/`Cstruct.t` but I don't see a real usage on my case.

> hmm, is there any plan for Cstruct_cap and these helpers? or is Cstruct_cap an interface to be discarded even before being widely used?

Due to the include, I think I can take a time to expose these functions into `Cstruct_cap` too (and it should help us to __use__ `Cstruct_cap` 👍 ). Let me come back with that.

- [ ] integrate it into `Cstruct_cap` interface

dinosaure:
I updated the PR to expose functions into `Cstruct_cap` interface. I tried to avoid conflicts with #262 but it seems good for me - and, on other side, we should integrate #262 into master if @hannesm is happy with it.

dinosaure:
CI is happy, I think it's ready to merge.

dinosaure:
I let this PR opens for one week more to let people to catch this update. Then, I will merge it and cut a release :+1:.

dinosaure:
One week after, ready to merge and I will cut a release.

PR #278, opened 2020-09-14
Delete length function

dinosaure:
Delete the `length` function which leads incompatibility with several packages when we do:
```ocaml
let length = Cstruct.len x in
let open Cstruct in
...
```

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -569,7 +569,6 @@ let buffer ?(off= 0) ?len buffer =
   if off < 0 || len < 0 || off + len > buffer_len then invalid_arg "index out of bounds" ;
   of_bigarray ~off ~len buffer
 
-let length cs = len cs
 let start_pos { off; _ } = off
 let stop_pos { off; len; _ } = off + len
 
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -585,9 +585,6 @@ val start_pos : t -> int
 val stop_pos : t -> int
 (** [stop_pos cs] is [cs]'s stop position in the base {!Cstruct.buffer}. *)
 
-val length : t -> int
-(** [length cs] is the number of bytes in [cs]. *)
-
 val head : ?rev:bool -> t -> char option
 (** [head cs] is [Some (get cs h)] with [h = 0] if [rev = false] (default) or [h
    = length cs - 1] if [rev = true]. [None] is returned if [cs] is empty. *)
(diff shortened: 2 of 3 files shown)


samoht:
i don't understand why this is an issue?

dinosaure:
It worth to keeping compatibility with packages - and don't add `length` - instead to break all for the already existing function `len`. The [release](https://github.com/ocaml/opam-repository/pull/17168) highlighted a breaking on `tls` and `dns` for example just because they use `length` as a value combined to an `let open Cstruct`, so `Cstruct.length` shadows the previous value and we finish with a compilation error.

talex5:
It's pretty confusing that cstruct provides both `len` and `length` functions (and with different doc-strings!).

```ocaml
val length : t -> int
(** [length cs] is the number of bytes in [cs]. *)

val len: t -> int
(** Returns the length of the current cstruct view.  Note that this
    length is potentially smaller than the actual size of the underlying
    buffer, as the [sub] or [set_len] functions can construct a smaller view. *)
```

samoht:
I'm in favour of deprecating `len` and advertise `lenght` more.

avsm:
I'm the opposite; `Cstruct.len` has been there since forever, and it'll be a long and tedious path to deprecating it, for no clear gain. I suggest:

- bumping this release to cstruct 6.0.0, since we have added a ton of new functions
- merge the docstrings to make it clear that `len` and `length` are exactly the same function.  We can recommend `length` for new code.
- Deprecate Cstruct.len in a future release 7.0.0, with an indication it'll be dropped in 8.0.0. There's really no rush; a function that's an alias of another name costs us nothing.

hannesm:
sorry to join the party late, but if you consider a 6.0.0, what about #262? from my point of view, the cstruct_cap is not very usable without the extensions in #262.... sorry, I've not much time to look into more detail / conflicts this week, but could do so next week.

dinosaure:
Agreed with @avsm including #262 as Hannes said. I can be in charge to do that properly.

avsm:
Version bumps are cheap :-) If you don't want to rush into the review of #262, we can just bump to 7.0.0 right afterwards for the Cstruct_cap change too and make sure that it works well.  I prefer not to load too many breaking changes into one release.   No strong opinion on this though.

hannesm:
@avsm I authored #262 ;) and feel it is stuck for unknown reasons in the merge chain... it even is approved by the cstruct_cap author :D

avsm:
I know - but there's a delay, just go ahead and cut 6.0 and then bump to 7.0 for #262. Don't feel pressure to merge _due to_ incoming version bumps, but of course, feel free to unblock long running patches because they need to get resolved! :)

dinosaure:
Close by #279

PR #279, opened 2020-09-14
Use Cstruct.length instead Cstruct.len (deprecation)

dinosaure:
In response to #278, this PR use only `Cstruct.length` and deprecate `Cstruct.len` (and fix the documentation). A question still exists: should we rename `lenv` to `lengthv`?

### DIFF
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -586,7 +586,9 @@ val stop_pos : t -> int
 (** [stop_pos cs] is [cs]'s stop position in the base {!Cstruct.buffer}. *)
 
 val length : t -> int
-(** [length cs] is the number of bytes in [cs]. *)
+(** Returns the length of the current cstruct view.  Note that this
+    length is potentially smaller than the actual size of the underlying
+    buffer, as the [sub] or [set_len] functions can construct a smaller view. *)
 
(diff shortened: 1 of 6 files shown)


lib/cstruct.mli
```diff
@@ -326,7 +326,7 @@ val blit_to_string: t -> int -> bytes -> int -> int -> unit
 val memset: t -> int -> unit
 (** [memset t x] sets all the bytes of [t] to [x land 0xff]. *)
 
-val len: t -> int
+val len: t -> int [@@deprecated "[since 2020-09] use [length] instead"]
```
hannesm, line 329:
```suggestion
val len: t -> int [@@deprecated "This function will be removed in cstruct 7.0.0. Please use [length] instead."]
```

dinosaure:
I think all people are agree with this PR and ready to a release of `cstruct.6.0.0`. I will cut a release tomorrow.

avsm:
Let's not deprecate Cstruct.len in this release please.  It's better to have both functions in one release, and then migrate code, and then deprecate Cstruct.len after that.

dinosaure:
Ah, yes, I misunderstood your steps, so we will deprecated `len` for `cstruct.7.0.0`?

PR #280, opened 2020-09-15
use ppxlib

pveber:
This patch implements a switch to `ppxlib` instead of `ppx_tools_versioned` which [is going to be deprecated](https://discuss.ocaml.org/t/ocaml-migrate-parsetree-2-0-0/5991). An invocation of `dune runtest` returns errors which I believe are not related to the patch, but to the compiler:
image
The word "Error" is now written in bold red by the compiler, and the terminal special characters  used to express that show up in the diff.

In implementing the change, I followed the original approach of a global transformation, although it might be preferable to use context-free rules as provided in `ppxlib`. I could help with that too.
Also, in `ppx_test/errors/pp.ml` I replaced `Migrate_parsetree.Driver.run_main` by `Ppxlib.Driver.standalone`, without being absolutely sure they do the same thing.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,9 +37,27 @@ jobs:
         ocaml-version: ${{ matrix.ocaml-version }}
     - run: opam pin add -n .
     - name: Packages
-      run: opam depext -yt cstruct cstruct-sexp cstruct-unix cstruct-lwt ppx_cstruct
+      run: opam depext -yt cstruct cstruct-sexp cstruct-unix cstruct-lwt
     - name: Build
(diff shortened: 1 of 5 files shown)


avsm:
Exceptionally useful patch, thank you @pveber! I'll review this shortly.

avsm:
Bumping the minimum version of ppx_cstruct to >=4.07.0 is fine by me.

dinosaure:
Should we include this PR to the next release (I'm ok with that)?

dinosaure:
Let's include this PR into the next release.

Issue #282, opened 2020-10-14
Deprecated Cstruct.len


dinosaure:
For the next release of `cstruct`, we should put a deprecation on `Cstruct.len` (and use `Cstruct.length` instead).

PR #284, opened 2020-11-25
Add a deprecation on Cstruct.len (use Cstruct.length instead)

dinosaure:
According to #282 to be ready (and to not forget) for a possible next release.

### DIFF
--- a/lib/cstruct.mli
+++ b/lib/cstruct.mli
@@ -337,7 +337,7 @@ val blit_to_string: t -> int -> bytes -> int -> int -> unit
 val memset: t -> int -> unit
 (** [memset t x] sets all the bytes of [t] to [x land 0xff]. *)
 
-val len: t -> int
+val len: t -> int [@@deprecated "len is deprecated, you should use length instead."]
 (** Returns the length of the current cstruct view.  Note that this
     length is potentially smaller than the actual size of the underlying
     buffer, as the [sub] or [set_len] functions can construct a smaller view. *)


hannesm:
is the predicate `deprecated` or `ocaml.deprecated`? i.e. does the above result in a deprecation warning? (I so far used `@@ocaml.deprecated`).

dinosaure:
According to [the official documentation](https://caml.inria.fr/pub/docs/manual-ocaml/attributes.html#ss%3Abuiltin-attributes), both can be used, but I can prepend with `ocaml` if you want.

hannesm:
@dinosaure thanks, I tested this locally, and it works great. I enabled OCaml-CI on this repository. Good to merge once CI (should we remove travis?) are finished.

hannesm (approved):

dinosaure:
The CI fails on the linter about `ppx/dune` which is not the part of this PR but the linter was running on this file in #285. To avoid conflict, I will not push a pass of `ocamlformat` on this file.

About Travis CI, I prefer to be conservative on CIs or we should open an issue to discuss about that.

hannesm:
> travis

I read this morning https://github.com/ocaml/ocaml-ci-scripts/issues/353 -- and I don't see a good reason to use travis in this repository (given that github actions and ocaml-ci are doing a good job :)

> The CI fails on the linter about ppx/dune which is not the part of this PR but the linter was running on this file in #285.

fine with me to fix that CI issue in a different PR :) good to merge this PR from my side.

PR #285, opened 2020-11-25
Remove color from ppx's binary to replicate diff of errors in any contexts

### LINKED ISSUE #283: ANSI escape sequences mess up ppx_test
`ppx_test` will fail with diffs like this:

```
--- ppx_test/errors/cstruct_len_not_int.ml.expected     2020-10-25 18:29:59.427677585 +0000
+++ ppx_test/errors/cstruct_len_not_int.ml.errors       2020-10-25 18:30:01.067662542 +0000
@@ -1 +1 @@
-Error: ppx_cstruct: [@len] argument should be an integer
+Error: ppx_cstruct: [@len] argument should be an integer
```

where the lower `Error:` would be colored red and the upper one not.

A workaround for this I have found is using `env TERM=dumb dune runtest -p ppx_cstruct`, but I suppose this issue should be prevented somehow in the first place.

dinosaure:
Try to fix #283 about `ppx_cstruct`, we ensure to never outputs with color and, by this way, be safe when we compare expected outputs and what the PPX tool give to us.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,2 @@
+version=0.15.0
+disable=true
--- a/ppx/dune
+++ b/ppx/dune
@@ -4,5 +4,6 @@
  (kind ppx_rewriter)
  (wrapped false)
  (ppx_runtime_libraries cstruct stdlib-shims)
- (preprocess (pps ppxlib.metaquot))
+ (preprocess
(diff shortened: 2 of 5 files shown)


sternenseemann:
Can confirm this patch fixes the issue I had with ppx_cstruct 6.0.0.

PR #286, opened 2021-03-09
Use Bytes_val if available

hannesm:
Since OCaml 4.06.0, String_val (when safe-string is enabled globally) does a
(const char *) cast. Here, we memcpy into a bytes buffer, and using String_val
emits a C compiler warning ("passing 'const char *' to parameter of type 'void *'
discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]").

To avoid this warning, we use Bytes_val if available (since 4.06.0).

### DIFF
--- a/lib/cstruct_stubs.c
+++ b/lib/cstruct_stubs.c
@@ -23,10 +23,14 @@
 #include <caml/alloc.h>
 #include <caml/bigarray.h>
 
+#ifndef Bytes_val
+#define Bytes_val String_val
+#endif
+
 CAMLprim value
 caml_blit_bigstring_to_string(value val_buf1, value val_ofs1, value val_buf2, value val_ofs2, value val_len)
 {
-  memcpy(String_val(val_buf2) + Long_val(val_ofs2),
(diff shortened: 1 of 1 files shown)


dinosaure (approved):

dinosaure:
> Not aware of any reason to keep newer cstruct supported on 4.03-4.06.

I'm not really sure about that, currently and according to my knowledge of the MirageOS eco-system, I try to keep such library (`cstruct` or `base64`) for `4.03.0` mostly because they are widely used by many users who don't try to follow like us the support of the 3 latest version of OCaml.

In my perspective, it's "free" to continue to support 4.03.0 for `cstruct` to give an opportunity to others libraries (which depends on `cstruct`) to stay compatible with 4.03.0. Then, specifically to this PR, we can do something like `base64` to support all versions. WDYT? (I can take care about the patch and the minor release of `cstruct` with this PR).

dinosaure:
I rebased the PR to master to include the work of @talex5 on the CI. I will merge (when the CI will be green) and cut a release then.

PR #287, opened 2021-06-23
Add shiftv

talex5:
This is particularly useful for advancing an iovec.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,5 +16,6 @@ steps:
       - opam pin add -n .
       - opam depext -u cstruct-async cstruct-lwt cstruct-unix cstruct ppx_cstruct
       - opam install -y .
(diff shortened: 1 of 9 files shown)


dinosaure (approved):

avsm (approved):
Looks right to me. @djs55 you had dstruct or something similar to work with lists of cstructs didn't you? Does this overlap with that?

talex5:
CI was failing (on master too). I've updated the opam metadata... I think I've found the magic combination that makes all the CIs happy!

dinosaure:
I'm ok to merge, do you want a release soon?

talex5:
A release would be nice at some point, but for now I've just inlined a copy in my code.

dinosaure:
Ok, will do a minor one tomorrow :+1:.

talex5:
Thanks!

PR #289, opened 2021-07-06
Update tests to use an official bun version

talex5:
Previously we pinned bun to a branch in my fork. The main repository now has the changes, and I'll make a new bun release soon.

This updates the drone tests to use the official repository for now, to test the latest code. I'll then update this PR to use opam-repository once the new bun is released.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,15 +7,12 @@ platform:
 
 steps:
   - name: build
-    image: ocaml/opam2:4.08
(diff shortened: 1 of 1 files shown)


avsm (approved):

talex5:
I've updated it to use the new release now. I also noticed we were still using the old `ocaml/opam2:4.08` image, so I also switched it to `ocaml/opam:debian-ocaml-4.12-afl`.

PR #293, opened 2022-02-28
Upgrade ocamlformat

dinosaure:


dinosaure:
> I don't understand the implications to approve this though, so better if someone else reviews.

That's fine 👍

PR #294, opened 2022-03-02
Be compatible with OCaml 5.00.0

dinosaure:
Add a new dependency on `cstruct-unix`: `mmap.1.2.0` to be compatible with OCaml 5.00.0 (/cc @kit-ty-kate).


dinosaure:
> Out of curiosity, what changed in 5.0, or why is Unix.map_file not enough anymore ?

This patch is needed due to: https://github.com/ocaml/dune/issues/5494

For a long support (including 4.05), we need to use `bigarray-compat` because `bigarray` (for OCaml <= 4.07) brings the `unix` dependency. So for a long term support, it's preferable to use `bigarray-compat` (even if, in that specific case, `bigarray`/`bigarray-compat` is not problem because the package - `cstruct-unix` - already depends on `unix`). Then, about OCaml 5.00, as I linked above, `bigarray` will be deleted (as a part of the caml distribution). It's a long run when we decided, may be 3 years ago, to integrate `bigarray` into the caml distribution **without** `map_file` which is now provided by `mmap` (for MirageOS) or `unix.cmxa` (for OCaml).

Due to the fact that `cstruct` is widely used, I prefer to keep a compatibility with, at least, 4.05.

haesbaert:
> > Out of curiosity, what changed in 5.0, or why is Unix.map_file not enough anymore ?
> 
> This patch is needed due to: [ocaml/dune#5494](https://github.com/ocaml/dune/issues/5494)
> 
> For a long support (including 4.05), we need to use `bigarray-compat` because `bigarray` (for OCaml <= 4.07) brings the `unix` dependency. So for a long term support, it's preferable to use `bigarray-compat` (even if, in that specific case, `bigarray`/`bigarray-compat` is not problem because the package - `cstruct-unix` - already depends on `unix`). Then, about OCaml 5.00, as I linked above, `bigarray` will be deleted (as a part of the caml distribution). It's a long run when we decided, may be 3 years ago, to integrate `bigarray` into the caml distribution **without** `map_file` which is now provided by `mmap` (for MirageOS) or `unix.cmxa` (for OCaml).
> 
> Due to the fact that `cstruct` is widely used, I prefer to keep a compatibility with, at least, 4.05.

Thank you for the explanation

patricoferris:
Thanks for this @dinosaure -- looks like `ppx_cstruct` could use a similar `bigarray` library removal

https://github.com/mirage/ocaml-cstruct/blob/554d411757041a76a2a3e01c33b97796b4a04d8d/ppx/dune#L9

dinosaure:
Indeed, I don't understand why we need such library for `ppx`.

hannesm:
I'd personally be also ok with raising the lower bound to 4.08.0 and removing all these `-compat` libraries (result, bigarray-compat, mmap) from our dependency chain. If someone is interested in earlier OCaml compilers, they can install old cstruct packages. The advantage of 4.08+ is as well e.g. the `Result` module (and let* syntax), which allows to remove the `rresult` dependency and write nice and readable code.

Given that OCaml 5.0.0 will have an impact on the ecosystem, I'd be in favour to drop the legacy boilerplate rather sooner than later to focus with a clear picture on the new major release (which may result in dropping even further legacy). Another reason is CI time / how many cycles are spent on a library. Given that both here and in opam-repository the CI attempts to compiler for all supported OCaml compilers, dropping support for old compilers is beneficial for the environment.

If we discover issues in cstruct we can always backport them to releases that work with OCaml before 4.08.

On a related note, since cmdliner 1.1.0 is supporting only OCaml >= 4.08.0, I can see that a huge part of our ecosystem is moving there already.

dinosaure:
> On a related note, since cmdliner 1.1.0 is supporting only OCaml >= 4.08.0, I can see that a huge part of our ecosystem is moving there already.

It seems a good argument indeed. I think the best is to see the intersection between:
- `cstruct >= 6.0.0`
- **and** `ocaml < 4.08`

If we don't find any packages inside this intersection, we can safely drop the support of `ocaml < 4.08`. I will try to check that and come back later in this PR - may be the best is to merge this PR and, if we want to drop the support of `ocaml < 4.08`, propose an other PR which remove this support.

hannesm:
@dinosaure sounds good. A quick `git grep cstruct | grep '>= "6.0.0"' | cut -d ':' -f 1 | xargs grep '"ocaml"' | grep -v '4.08' | less` in opam-repository doesn't show anything that depends on `cstruct >= 6.0.0` and requires ocaml < 4.08.0. The above output: (a) everything mentioning 4.08.0 is `>= 4.08.0`, (b) is non-empty, some packages requires earlier OCaml versions, other newer ones than 4.08.0. TL;DR: IMHO all safe to raise the lower bound to 4.08.0.

PR #295, opened 2022-03-02
Lint dependencies and the CI

dinosaure:


dinosaure:
It seems clear that we must drop the support of `< 4.05.0` due to dependencies. `cstruct` is the package which as the longer support but I think it's probably fine to move on when, for any MirageOS projects, we support, at least, the three latest version of OCaml.

dinosaure:
The error on windows is unrelated (it's about `lwt` on Windows). I think it's good to merge and prepare a release then.

PR #297, opened 2022-03-14
CHANGES.md: fix year

reynir:
I was confused why 6.0.1 was released three months earlier than 6.0.0 ;-)


dinosaure:
Yes, I probably did a mistake :+1:. Thanks.

PR #298, opened 2022-03-22
drop support for ocaml < 4.08.0, drop bigarray-compat dependency

hannesm:
as discussed in #296

PR #300, opened 2022-07-04
ppx_cstruct: remove stdlib-shims dependency, it requires OCaml 4.08 already

hannesm:
somehow overlooked in #298

### DIFF
--- a/ppx/dune
+++ b/ppx/dune
@@ -3,7 +3,7 @@
  (public_name ppx_cstruct)
  (kind ppx_rewriter)
  (wrapped false)
- (ppx_runtime_libraries cstruct stdlib-shims)
+ (ppx_runtime_libraries cstruct)
  (preprocess
   (pps ppxlib.metaquot))
- (libraries sexplib ppxlib stdlib-shims))
+ (libraries sexplib ppxlib))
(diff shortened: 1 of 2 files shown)


hannesm:
//cc @dinosaure

hannesm:
any idea what the "drone" CI issue is? it reports that some crashes have been found, but I'm not able to reproduce. any help?

samoht:
```
17:19.18:[ERROR]All fuzzers finished, but some crashes were found!
```

That's indeed not super helpful!

dinosaure (approved):
I will try to take a look tomorrow.

dinosaure:
The error seems related to an OOM. It's reproducible on my machine if I do:
```sh
$ cd _build/default/fuzz
$ afl-fuzz -m 200 -i input -o output -S 3 -- ./fuzz.exe @@
```

I'm not sure how to fix that and even if I put something larger than 200 (with the `-m` option), the fuzzer continue to fail. It's probably an error into the `fuzz.exe` than the `afl-fuzz` - may be, it allocates a lot. In any way, I will make an issue about that and merge this PR when the problem seems unrelated.

PR #302, opened 2022-10-26
unix: add Cstruct_unix.{read,write,writev,send,recv}

djs55:
This is useful for direct-style code to avoid copying Cstructs into temporary byte buffers to use the `Unix.` API.

Signed-off-by: David Scott <dave@recoil.org>

### DIFF
--- a/unix/read_stubs.c
+++ b/unix/read_stubs.c
@@ -1,0 +1,70 @@
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+#include <caml/custom.h>
+#include <caml/callback.h>
+#include <caml/alloc.h>
+#include <caml/unixsupport.h>
+#include <caml/bigarray.h>
+#include <caml/threads.h>
+
+#include <stdio.h>
+#include <errno.h>
+
+CAMLprim value stub_cstruct_read(value val_fd, value val_c)
+{
+  CAMLparam2(val_fd, val_c);
+  CAMLlocal3(val_buf, val_ofs, val_len);
+
+  val_buf = Field(val_c, 0);
+  val_ofs = Field(val_c, 1);
+  val_len = Field(val_c, 2);
+
+  void *buf = (char *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+  size_t len = Long_val(val_len);
+  int n = 0;
+
+#ifdef _WIN32
+  int win32err = 0;
+  switch (Descr_kind_val(val_fd))
+  {
+  case KIND_SOCKET:
+    SOCKET s = Socket_val(val_fd);
+
+    caml_release_runtime_system();
+    n = recv(s, buf, len, 0);
+    win32err = WSAGetLastError();
+    caml_acquire_runtime_system();
+
+    if (n == SOCKET_ERROR)
+    {
+      win32_maperr(win32err);
+      unix_error(errno, "read", Nothing);
+    }
+    break;
+  case KIND_HANDLE:
+    HANDLE h = Handle_val(val_fd);
+    DWORD numread;
+    caml_release_runtime_system();
+    int ok = ReadFile(h, buf, len, &numread, NULL);
+    win32err = GetLastError();
+    n = numread;
+    caml_acquire_runtime_system();
+
+    if (!ok)
+    {
+      win32_maperr(win32err);
+      unix_error(errno, "read", Nothing);
+    }
+    break;
+  default:
+    caml_failwith("unknown Descr_kind_val");
+  }
+#else
+  caml_release_runtime_system();
+  n = read(Int_val(val_fd), buf, len);
+  caml_acquire_runtime_system();
+  if (n < 0)
+    unix_error(errno, "read", Nothing);
+#endif
+  CAMLreturn(Val_int(n));
+}
--- a/unix/recv_stubs.c
+++ b/unix/recv_stubs.c
@@ -1,0 +1,63 @@
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+#include <caml/custom.h>
+#include <caml/callback.h>
+#include <caml/alloc.h>
+#include <caml/unixsupport.h>
+#include <caml/bigarray.h>
+#include <caml/threads.h>
+
+#include <stdio.h>
+
+#ifdef WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#include <NTSecAPI.h>
+#else
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <errno.h>
+#endif
+
+CAMLprim value stub_cstruct_recv(value val_fd, value val_c)
+{
+    CAMLparam2(val_fd, val_c);
+    CAMLlocal3(val_buf, val_ofs, val_len);
+
+    val_buf = Field(val_c, 0);
+    val_ofs = Field(val_c, 1);
+    val_len = Field(val_c, 2);
+
+    void *buf = (void *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+    size_t len = (size_t)Long_val(val_len);
+    int n = 0;
+#ifdef WIN32
+    int win32err = 0;
+    if (Descr_kind_val(val_fd) != KIND_SOCKET)
+        unix_error(EINVAL, "recv", Nothing);
+
+    SOCKET s = Socket_val(val_fd);
+
+    caml_release_runtime_system();
+    n = recv(s, buf, len, 0);
+    win32err = WSAGetLastError();
+    caml_acquire_runtime_system();
+
+    if (n == SOCKET_ERROR)
+    {
+        win32_maperr(win32err);
+        unix_error(errno, "recv", Nothing);
+    }
+#else
+    int fd = Int_val(val_fd);
+
+    caml_release_runtime_system();
+    n = recv(fd, buf, len, 0);
+    caml_acquire_runtime_system();
+
+    if (n < 0)
+        unix_error(errno, "recv", Nothing);
+#endif
+    CAMLreturn(Val_int(n));
+}
--- a/unix/send_stubs.c
+++ b/unix/send_stubs.c
@@ -1,0 +1,62 @@
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+#include <caml/custom.h>
+#include <caml/callback.h>
+#include <caml/alloc.h>
+#include <caml/unixsupport.h>
+#include <caml/bigarray.h>
+#include <caml/threads.h>
+
+#include <stdio.h>
+
+#ifdef WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <winsock2.h>
+#include <ws2tcpip.h>
+#include <NTSecAPI.h>
+#else
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <errno.h>
+#endif
+
+CAMLprim value stub_cstruct_send(value val_fd, value val_c)
+{
+    CAMLparam2(val_fd, val_c);
+    CAMLlocal3(val_buf, val_ofs, val_len);
+
+    val_buf = Field(val_c, 0);
+    val_ofs = Field(val_c, 1);
+    val_len = Field(val_c, 2);
+
+    const char *buf = (char *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+    size_t len = (size_t)Long_val(val_len);
+    int n = 0;
+
+#ifdef WIN32
+    int win32err = 0;
+    if (Descr_kind_val(val_fd) != KIND_SOCKET)
+        unix_error(EINVAL, "send", Nothing);
+
+    SOCKET s = Socket_val(val_fd);
+    caml_release_runtime_system();
+    n = send(s, buf, len, 0);
+    win32err = WSAGetLastError();
+    caml_acquire_runtime_system();
+
+    if (n == SOCKET_ERROR)
+    {
+        win32_maperr(win32err);
+        unix_error(errno, "send", Nothing);
+    }
+#else
+    int fd = Int_val(val_fd);
+
+    caml_release_runtime_system();
+    n = send(fd, buf, len, 0);
+    caml_acquire_runtime_system();
+    if (n < 0)
+        unix_error(errno, "send", Nothing);
+#endif
+    CAMLreturn(Val_int(n));
+}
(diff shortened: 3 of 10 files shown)


haesbaert:
Me gusta ! I've been carrying this around for ages: https://github.com/haesbaert/rawlink/blob/520207190814ebfc0db9fd4f6d9c6e0f1297732d/lib/rawlink_stubs.c#L466

unix/recv_stubs.c
```diff
@@ -0,0 +1,63 @@
...
+    val_ofs = Field(val_c, 1);
+    val_len = Field(val_c, 2);
+
+    void *buf = (void *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+    size_t len = (size_t)Long_val(val_len);
+    int n = 0;
+#ifdef WIN32
+    int win32err = 0;
+    if (Descr_kind_val(val_fd) != KIND_SOCKET)
+        unix_error(EINVAL, "recv", Nothing);
```
avsm, line 38:
```suggestion
        unix_error(EINVAL, "stub_cstruct_recv", Nothing);
```

makes it easier to grep for the source of the error if we just name it after the function.

unix/unix_cstruct.ml
```diff
@@ -17,3 +17,60 @@
...
+    write fd @@ Cstruct.shift buf n
+  end
+
+let writev fd bufs =
+  let rec use_writev = function
+    | [] -> ()
+    | remaining ->
+      (* write at most iov_max at a time *)
+      let to_send = first iov_max [] remaining in
+      let n = stub_writev fd (List.map (fun x -> x.Cstruct.buffer, x.Cstruct.off, x.Cstruct.len) to_send) in
```
avsm, line 60: do you expand to a tuple here because the Cstruct.t record type might change? Given this is in the same repository, it's probably safe enough to just pass the Cstruct record in directly to the FFI and save some allocations

unix/recv_stubs.c
```diff
@@ -0,0 +1,63 @@
...
+CAMLprim value stub_cstruct_recv(value val_fd, value val_c)
+{
+    CAMLparam2(val_fd, val_c);
+    CAMLlocal3(val_buf, val_ofs, val_len);
+
+    val_buf = Field(val_c, 0);
+    val_ofs = Field(val_c, 1);
+    val_len = Field(val_c, 2);
+
+    void *buf = (void *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
```
haesbaert, line 32: minor bikeshedding, would move `buf` to the beginning of the block, declarations after statement is not valid C89 (not sure if anyone cares these days)
Also void * arithmetic is technically undefined behaviour.

unix/send_stubs.c
```diff
@@ -0,0 +1,62 @@
...
+    int n = 0;
+
+#ifdef WIN32
+    int win32err = 0;
+    if (Descr_kind_val(val_fd) != KIND_SOCKET)
+        unix_error(EINVAL, "send", Nothing);
+
+    SOCKET s = Socket_val(val_fd);
+    caml_release_runtime_system();
+    n = send(s, buf, len, 0);
```
haesbaert, line 43: same as recv

unix/write_stubs.c
```diff
@@ -0,0 +1,69 @@
...
+CAMLprim value stub_cstruct_write(value val_fd, value val_c)
+{
+  CAMLparam2(val_fd, val_c);
+  CAMLlocal3(val_buf, val_ofs, val_len);
+  val_buf = Field(val_c, 0);
+  val_ofs = Field(val_c, 1);
+  val_len = Field(val_c, 2);
+  void *buf = (char *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+  size_t len = Long_val(val_len);
+  int n = 0;
```
haesbaert, line 22: should be ssize_t

unix/writev_stubs.c
```diff
@@ -0,0 +1,69 @@
...
+#include <caml/threads.h>
+
+#include <stdio.h>
+#include <errno.h>
+#ifndef _WIN32
+#include <sys/uio.h>
+#include <limits.h>
+#endif
+
+CAMLprim
```
haesbaert, line 17: bikeshedding, the other stubs don't break line here

unix/read_stubs.c
```diff
@@ -0,0 +1,70 @@
...
+#include <caml/callback.h>
+#include <caml/alloc.h>
+#include <caml/unixsupport.h>
+#include <caml/bigarray.h>
+#include <caml/threads.h>
+
+#include <stdio.h>
+#include <errno.h>
+
+CAMLprim value stub_cstruct_read(value val_fd, value val_c)
```
haesbaert, line 13: Can this be marked as noalloc ? I think unix_error allocs !?

avsm: definitely can't be noalloc -- it releases the runtime lock as well.

unix/unix_cstruct.ml
```diff
@@ -17,3 +17,60 @@
 let of_fd fd =
   let buffer = Bigarray.(array1_of_genarray (Unix.map_file fd char c_layout false [|-1|])) in
   Cstruct.of_bigarray buffer
+
+type buffer = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
```
dinosaure, line 21: You can re-use the type `Cstruct.buffer` here.

unix/recv_stubs.c
dinosaure: I think `void *` is used here to be in-sync with what `recv` expects:
```c
ssize_t recv(int sockfd, void *buf, size_t len, int flags);
```

haesbaert: Ack but passing pointer of any type to something that expects a void pointer is ok, my only bit was that we're doing (void *)++, which is undefined behavior (but every compiler implicitly assumes char *) so it's ok.

haesbaert:
I can provide Cstruct_unix.recvfrom and Cstruct_unix.sendto if it's desirable, I just can't do the windows bits.

unix/recv_stubs.c
talex5: https://stackoverflow.com/questions/3922958/void-arithmetic says it's a GCC extension, and attempting this should produce warnings.

unix/writev_stubs.c
```diff
@@ -0,0 +1,69 @@
+#include <caml/mlvalues.h>
```
haesbaert, line 1:
```suggestion
#define _XOPEN_SOURCE /* IOV_MAX */
#include <caml/mlvalues.h>
```
Linux doesn't export IOV_MAX unless you define _XOPEN_SOURCE or _GNU_SOURCE

unix/unix_cstruct.ml
djs55: Yes that was it -- I was using these stubs in another repo. I'll drop the tuple since we're here.

unix/unix_cstruct.mli
```diff
@@ -39,3 +39,9 @@ val send: Unix.file_descr -> Cstruct.t -> int
 val recv: Unix.file_descr -> Cstruct.t -> int
 (** [recv fd c] receives a message from a socket. This can be used to receive a datagram.
     If only a partial receive is possible, the return argument is now many bytes were received. *)
+
+val recvfrom: Unix.file_descr -> Cstruct.t -> Unix.msg_flag list -> int * Unix.sockaddr
+(** [recvfrom fd c] Like Unix.recvfrom, but for Cstruct. *)
```
avsm, line 44:
```suggestion
(** [recvfrom fd c] Like {! Unix.recvfrom}, but for Cstruct. *)
```

will link to it on the ocaml.org docs then.

unix/recvfrom_stubs.c
```diff
@@ -0,0 +1,50 @@
...
+  buf = (uint8_t *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
+  len = Long_val(val_len);
+  addr_len = sizeof(addr);
+
+  caml_release_runtime_system();
+  n = recvfrom(Int_val(val_fd), buf, len, cv_flags, &addr.s_gen, &addr_len);
+  caml_acquire_runtime_system();
+
+  if (n == -1)
+    caml_uerror("recvfrom", Nothing);
```
avsm, line 42:
```suggestion
    caml_uerror("cstruct_recvfrom", Nothing);
```

I do find these Unix_errors are marginally more useful if you can grep for the function they came from

unix/recv_stubs.c
```diff
@@ -0,0 +1,63 @@
...
+    if (n == SOCKET_ERROR)
+    {
+        win32_maperr(win32err);
+        unix_error(errno, "recv", Nothing);
+    }
+#else
+    int fd = Int_val(val_fd);
+
+    caml_release_runtime_system();
+    n = recv(fd, buf, len, 0);
```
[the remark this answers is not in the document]
djs55: Good point, I’ve copied the flags code from your stubs :)

unix/recv_stubs.c
```diff
@@ -0,0 +1,70 @@
...
+#include <ws2tcpip.h>
+#include <NTSecAPI.h>
+#else
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <errno.h>
+#endif
+
+static int msg_flag_table[] = {
+    MSG_OOB, MSG_DONTROUTE, MSG_PEEK /* XXX */
```
avsm, line 24: what's the XXX for here? Incomplete flags?

unix/unix_cstruct.ml
```diff
@@ -17,3 +17,56 @@
...
+external stub_iov_max: unit -> int = "stub_cstruct_iov_max"
+
+external stub_write: Unix.file_descr -> Cstruct.t -> int = "stub_cstruct_write"
+
+external stub_writev: Unix.file_descr -> Cstruct.t list -> int = "stub_cstruct_writev"
+
+let iov_max = stub_iov_max ()
+
+(* return the first n fragments, suitable for writev *)
+let rec first n rev_acc = function
```
avsm, line 31: I mildly dislike having to do these potentially O(n) list operations for a writev, and mandate that in the external interface.   In your vpnkit usecase, would you be able to pass in a `Cstruct.t array` instead for writev, and then just call writev with simple O(1) time slices instead?

It affects the interface we expose in the mli, hence my asking now (as opposed to optimising later)

unix/recv_stubs.c
haesbaert: Cause we assume how stdlib ordered things and also because we redefine them, would be nice if stdlib exposed them as not static so we could just use caml_msg_flag_table or something

unix/write_stubs.c
```diff
@@ -0,0 +1,69 @@
...
+#include <errno.h>
+
+CAMLprim value stub_cstruct_write(value val_fd, value val_c)
+{
+  CAMLparam2(val_fd, val_c);
+  CAMLlocal3(val_buf, val_ofs, val_len);
+  val_buf = Field(val_c, 0);
+  val_ofs = Field(val_c, 1);
+  val_len = Field(val_c, 2);
+  void *buf = (char *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
```
haesbaert, line 20:
```suggestion
  uint8_t *buf = (uint8_t *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
```

unix/writev_stubs.c
haesbaert: Still breaking line here in stub_cstruct_iov_max

unix/writev_stubs.c
```diff
@@ -0,0 +1,68 @@
...
+  int c_fd = Int_val(fd);
+
+  next = val_list;
+  /* Calculate the length of the val_list */
+  int length = 0;
+  for (next = val_list; next != Val_emptylist; next = Field(next, 1))
+    length++;
+  /* Only copy up to the iovec array size */
+  if (length > IOV_MAX)
+    length = IOV_MAX;
```
haesbaert, line 48: Silently truncating it might be worse than just letting the syscall failing later on in `uerror` with EINVAL.
You exposed the maximum, if the user goes above he should be punished with an exception imho.
It's a bit harmless for a STREAM that has to handle shortcounts, but a DATAGRAM will send half a packet.
(even if we're all a bit insane because IOV_MAX is huge)

unix/writev_stubs.c
```diff
@@ -0,0 +1,68 @@
...
+  next = val_list;
+  for (i = 0; i < length; i++)
+  {
+    head = Field(next, 0);
+    val_buf = Field(head, 0);
+    val_ofs = Field(head, 1);
+    val_len = Field(head, 2);
+    iovec[i].iov_base = (char *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
```
haesbaert, line 56:
```suggestion
    iovec[i].iov_base = (uint8_t *)Caml_ba_data_val(val_buf) + Long_val(val_ofs);
```

unix/recv_stubs.c
haesbaert: Ah that's smart ! yes we should do that.

unix/writev_stubs.c
avsm: We should mostly mirror what writev(2) itself does, and that returns an EINVAL if > IOV_MAX.  So I agree that a uerror should be raised if it's too large, and not a silent truncation.

unix/write_stubs.c
dinosaure: I approve this change, the FFI about `bigarray` says that it's an `uint8_t` array. We should keep this assumption on the C side too.

dinosaure:
A release should be done soon but this PR has some remaining issues. As far as I can say, it's mostly about style. @djs55 can you do a review of them?

patricoferris:
Just thought I would mention that I will try to get a `pread/pwrite/preadv/pwritev` PR open soonish. I started hacking on that over here https://github.com/mirage/ocaml-cstruct/commit/4fd8e394058b3a4a86d741e3d0766815569e8dd6 but don't currently have a windows machine to do that part and couldn't get macOS (11+) to be happy about the `preadv/pwritev` functions. I mention this because it would help with the potential IOCP/kqueue backends in [Eio](https://github.com/ocaml-multicore/eio). Just thought I would mention it in case others were already hacking on this :))

patricoferris:
I haven't worked out a nice story for supporting `pread/pwrite` on Windows because the calls to `ReadFile/WriteFile` with an [Overlapped](https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped) changes the current position (`pread/pwrite` do not change the position). Lwt_unix has decided to leave that as undefined behaviour (see https://github.com/ocsigen/lwt/pull/768). Would that be okay here?

dinosaure:
> Would that be okay here?

I think it's ok when the module explicitely mention `unix` in it's name 🙂 . A note (like `lwt`) will be cool to inform the user if he/she tries to use this module on Windows.

PR #308, opened 2023-02-10
Deprecate copy

reynir:
As briefly discussed in https://github.com/mirage/ocaml-cstruct/pull/305#issuecomment-1424637704 `Cstruct.copy` is basically `Cstruct.to_string` without defaults/labelled arguments. The name `copy` suggests it creates a cstruct from a cstruct, but it creates a string. Instead, the new `Cstruct.sub_copy` makes a cstruct copy of a cstruct. I propose we deprecate `Cstruct.copy` with a message suggesting to use either `Cstruct.to_string` or `Cstruct.sub_copy`, and some time in the future we may rename `Cstruct.sub_copy` to `Cstruct.copy`.

The ppx still uses names like `copy_struct_field`. It's unclear to me if we want to preserve this, or try to deprecate it in favor of `to_string_struct_field` (or similar).

### DIFF
--- a/lib/cstruct.ml
+++ b/lib/cstruct.ml
@@ -53,7 +53,7 @@ let err_of_bigarray t = err "Cstruct.of_bigarray off=%d len=%d" t
 let err_sub t = err "Cstruct.sub: %a off=%d len=%d" pp_t t
 let err_shift t = err "Cstruct.shift %a %d" pp_t t
 let err_shiftv n = err "Cstruct.shiftv short by %d" n
-let err_copy t = err "Cstruct.copy %a off=%d len=%d" pp_t t
+let err_copy_to_string caller t = err "Cstruct.%s %a off=%d len=%d" caller pp_t t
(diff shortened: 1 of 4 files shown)


dinosaure:
Can you extends that on the `cstruct_cap.mli` API too?

reynir:
Apparently `Cstruct_cap.copy` doesn't exist. Maybe we don't need to add it if we plan on deprecating it?

While looking for `Cstruct_cap.copy` I realize we have `Cstruct.copyv`...

dinosaure:
> Apparently Cstruct_cap.copy doesn't exist. Maybe we don't need to add it if we plan on deprecating it?

The deprecation of `Cstruct_cap` is orthogonal to this PR and I would like to keep in-sync `Cstruct` & `Cstruct_cap` even if we start to think the uselessness of `Cstruct_cap`.

> While looking for Cstruct_cap.copy I realize we have Cstruct.copyv...

Yes, I think a rational exists about the non-existence of `copy` when `to_string` (+ `blit`) do, in a certain way, the same job. The PR is good for me.

dinosaure:
Thanks!

PR #313, opened 2023-06-19
Rename Cstruct.lenv -> Cstruct.lengthv

reynir:
Since `Cstruct.len` was renamed to `Cstruct.length` it would make sense to apply the same renaming to `Cstruct.lenv`.

We may want to have a deprecation period instead.


avsm:
This really does need a compatibility shim for a few releases. These renames cause an awful lot of churn in downstream users of Cstruct, so giving them plenty of time to adapt is good.

Issue #315, opened 2023-07-27
Dependency on `ocaml-migrate-parsetree`


pitag-ha:
Hello :wave:

Is OMP still a test dependency of `ppx_cstruct`? It is according to its `opam` file (both opam-repo and local) and according to the test `dune` file. Looking at the test, it seems to use Ppxlib instead, though, doesn't it?

fogti:
Yeah, appears to work without it. https://github.com/gentoo/guru/pull/144/files#diff-d332b7c85532a6952cb892936b32297595aaf3df7cb02115388420d51c8bc5e1

avsm:
Getting rid of it in #325!

PR #316, opened 2024-09-09
Add sexp_of cenum attribute, use sexplib0

reynir:
The `sexp_of_*` functions now return `Sexplib0.Sexp.t` instead of `Sexplib.Sexp.t`. The `cenum` attribute `[@@sexp_of]` is like `[@@sexp]` except it only produces the `sexp_of*` function. This allows for generating `sexp_of` converters without depending on the full sexplib library.

### DIFF
--- a/ppx/ppx_cstruct.ml
+++ b/ppx/ppx_cstruct.ml
@@ -392,7 +392,7 @@ type cenum =
   { name : string Loc.t;
     fields : (string Loc.t * int64) list;
     prim : prim;
-    sexp : bool;
+    sexp : [ `None | `Sexp_of | `Sexp ];
   }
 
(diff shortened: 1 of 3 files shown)


reynir:
I found it wasn't too difficult to use sexplib0 instead of sexplib in cstruct-sexp so I went ahead with that. I think it might be possible to entirely move to sexplib0 over sexplib....

reynir:
I can't quite figure out macos CI. I note we use avsm/setup-ocaml@v1 -- should we use ocaml/setup-ocaml@v2 instead? And how to migrate?

hannesm:
taking #315 into account, to fix the CI here, maybe we should remove ocaml-migrate-parsetree dependency?

PR #320, opened 2026-07-07
cstruct: ship a Wasm runtime (cstruct.wat), the analog of cstruct.js

samoht:
wasm_of_ocaml resolves a library's C primitives from Wasm modules registered with (wasm_of_ocaml (wasm_files ...)), importing runtime helpers from the env namespace. cstruct.wat mirrors cstruct.js one to one: the three bigstring blits alias the runtime's caml_bigstring_blit_* exports, fill and compare loop over caml_ba_set_1 / caml_ba_get_1 + caml_int_compare, and check_alignment is always true. Without it a program that links cstruct cannot run as wasm (the blit primitives get dummy stubs that throw).


vouillon:
Looks good. The functions already exist under another name. So you can also just reexport them.
https://github.com/ocaml-wasm/ocaml-cstruct/blob/cc77042b9ce9b27f5f4ed2d18a705e1143da4951/lib/cstruct.wat

samoht:
@avsm any objection with shipping wasm files in there?

avsm:
No objection here. A test case would be nice if possible just to check the linkage is working but not a blocker

PR #322, opened 2026-07-07
Modernize the GitHub Actions CI

samoht:
CI currently fails on every PR: `avsm/setup-ocaml@v1` errors with "the architecture is not supported" on today's runners, so all jobs abort in ~10s. This switches to `ocaml/setup-ocaml@v3` + `actions/checkout@v4`, drops the removed `opam depext` step, and bumps the matrix to OCaml 5.3/5.1/4.14 (which ship a dune supporting the 3.17 language, needed once #321 lands).


samoht:
Need https://github.com/ocaml/opam-repository/pull/30217 to fix the ocaml-ci issues.

samoht:
@avsm I've removed the Drone CI as it seems to be completely dead since they got acquired.

avsm:
Looks good, thanks!

